PR #2, opened 2025-01-07
add method to remove .s from nodeName

omrishiv:
At the moment, it's not possible to set the pod.spec.hostName to a name with `.`s. This is a problem if the node name has a `.` in it, which is allowed. This PR addresses that by taking the first part of the node name before the `.` and setting that as the pod.spec.hostName

### DIFF
--- a/internal/controller/nodeset/nodeset_update.go
+++ b/internal/controller/nodeset/nodeset_update.go
@@ -368,7 +368,7 @@ func (nsc *defaultNodeSetControl) syncSlurm(
 		_ = slurmtypes.NodeInfoParse(node.Comment, &nodeInfo)
 		noPodInfo := nodeInfo.Equal(slurmtypes.NodeInfo{})
 		if kubeNodes.Has(node.Name) || !hasCommunicationFailure || noPodInfo {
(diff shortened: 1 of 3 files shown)


odellem:
This would split the name across the first `.` right? Have you tested if a name like `node.a.b` can be set with the entire name? The above example looks like it would split on the first period and set it as `node`

omrishiv:
We only want what's left of the first `.`. Any other ones are going to break the deployment of the compute group as you cannot have `.`s in the pod `hostName`

omrishiv:
Thinking about this more, we can have collisions if we only look at the first part. I'll update the code in a few hours and add some more tests.

marqvi:
Just run into the same issue with EKS node names. Would be very interested in the new nodeset controller when it's ready.

omrishiv:
I ran into a further issue scaling nodes which I also had to address. Is there a way to contribute these back or do we wait for a new nodeset release?

catblade:
We've pushed up the new node controller. @omrishiv  could you please check that out?

omrishiv:
Thanks @catblade, I’ll give it a shot

PR #3, opened 2025-01-10
nodeset: do not undrain node automatically

thomasbarrett:
When a node is drained within slurm, the slurm-operator will automatically undrain it unless the pod is marked for deletion. This behavior makes it very difficult to use a slinky cluster without administrative control over the host kubernetes cluster. It is impossible to prevent jobs from running on a (potentially broken) node from within the cluster.

I'm sure that there is a better way to do this, so I am super open to feedback on how this can be done.

### DIFF
--- a/internal/controller/nodeset/nodeset_control.go
+++ b/internal/controller/nodeset/nodeset_control.go
@@ -427,7 +427,7 @@ func (nsc *defaultNodeSetControl) processNodeSetPod(
 	}
 
 	stateMatch := true
-	if isNodeSetPodCordon(pods[i]) || nsc.podControl.isNodeSetPodDrain(ctx, set, pods[i]) {
+	if isNodeSetPodCordon(pods[i]) {
 		stateMatch = false
 	}
 
--- a/internal/controller/nodeset/nodeset_pod_control.go
+++ b/internal/controller/nodeset/nodeset_pod_control.go
@@ -196,35 +196,6 @@ func (spc *NodeSetPodControl) updateSlurmNode(
 		return err
 	}
 
-	if spc.isNodeSetPodDrain(ctx, set, pod) {
-		clusterName := types.NamespacedName{
-			Namespace: set.GetNamespace(),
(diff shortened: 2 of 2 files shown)


odellem:
Would you want a test for this case?

catblade:
We will release a new node controller in the next while, and will add this to the use case to our testing, probably a few months. Can you add this as an issue, so we can make sure to track it?

catblade:
Could you please add a full issue writeup here? Would appreciate that.

thomasbarrett:
This repo doesn't have public issues enabled, so I can't create an issue. I would love to help contribute to this project :)

thomasbarrett:
The issues is as follows:

When a node on a slurm cluster has broken hardware, the typical response on a normal slurm cluster is to drain the node. This does not work because the node controller will immediately undrain the node. It should be possible to drain nodes within the cluster.

SkylerMalinowski:
> When a node is drained within slurm, the slurm-operator will automatically undrain it unless the pod is marked for deletion. This behavior makes it very difficult to use a slinky cluster without administrative control over the host kubernetes cluster. It is impossible to prevent jobs from running on a (potentially broken) node from within the cluster.

We are aware of this operational gap and will be addressing it in the next major release (v0.2.0). I cannot commit to a release time frame yet.

Due to the large nodeset controller refactor and changes slated, this PR will become incompatible and obsolete. I will close this PR in the meantime. Thank you for your understanding and patience. :slightly_smiling_face: 

> I would love to help contribute to this project. How can I get involved[?]

As of right now: reaching out to <slinky@schedmd.com> or <sales@schemd.com>; opening PRs/MRs; chatting with us on Slack (CNCF, Kubernetes).

SkylerMalinowski:
FYI: On main, commit a8caac14e77ec42fade256f2e8ed7d808cf56619 has solved this issue. The NodeSet controller code has changed significantly (main compared to release-0.1 branch).

PR #7, opened 2025-03-04
Minor chart fixes

kincl:
The first commit is just about being more explicit with requirements:
* NET_RAW is required for using ping 
* Finalizers RBAC is required when setting OwnerReferences if the Kubernetes cluster has the [OwnerReferencesPermissionEnforcement](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#ownerreferencespermissionenforcement) admission controller plugin enabled

Second commit disables unshare in slurmrestd which does not work on unprivileged containers

### DIFF
--- a/helm/slurm/templates/cluster/token-job-rbac.yaml
+++ b/helm/slurm/templates/cluster/token-job-rbac.yaml
@@ -22,9 +22,11 @@ rules:
   - ""
   resources:
   - secrets
+  - secrets/finalizers
   verbs:
   - create
   - get
+  - update
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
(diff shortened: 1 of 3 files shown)


helm/slurm/templates/cluster/token-job-rbac.yaml
```diff
@@ -22,9 +22,11 @@ rules:
   - ""
   resources:
   - secrets
+  - secrets/finalizers
   verbs:
   - create
   - get
+  - update
```
SkylerMalinowski, line 29: It is not clear why this verb was added/needed. Can you please elaborate on this?

SkylerMalinowski:
Merged in: https://github.com/SlinkyProject/slurm-operator/compare/692cd35694e80fec3a3a249bd526c6ae5bab94c4~..5d311c80422da7d821bf76e77a32cd926ddfd45c

PR #10, opened 2025-04-18
feat: match Slurm resources with Kubernetes for CPU/Memory

kincl:
Adds support for mirroring the Kubernetes container resource request to slurmd. 

Kubernetes accepts a [wide array of options for memory resource units ](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-memory) which I implemented except for "milibytes" with lowercase m.

### DIFF
--- a/helm/slurm/templates/_helpers.tpl
+++ b/helm/slurm/templates/_helpers.tpl
@@ -83,3 +83,86 @@ Define image tag
 {{- define "slurm.image.tag" -}}
 {{- printf "%s-ubuntu24.04" .Chart.AppVersion -}}
 {{- end }}
+
+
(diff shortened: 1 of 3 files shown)


kincl:
In doing some testing here, it looks like the features don't get replaced once the node is defined? So if disable this feature at first and then add it in later I cannot change what the database has for the node's CPUs?

SkylerMalinowski:
FYI: this concept has been implemented as follows:
https://github.com/SlinkyProject/slurm-operator/commit/3e63fef9807c1775db9c3d20efd306069ac13a13
https://github.com/SlinkyProject/containers/commit/2e0c8d95f60db65e083e26cd21e6de3c02689119

PR #13, opened 2025-05-26
Add configmap to compute nodeset

faganihajizada:
## Problem
GPU nodes were showing as `invalid` state due to missing `gres.conf` configuration file in compute pods. The issue was that the `slurm-config` projected volume in the compute nodeset template only included the secret but not the configmap containing configuration files.

## Root Cause
In `helm/slurm/templates/compute/compute-nodeset.yaml`, the `slurm-config` volume was missing the configmap source:

```yaml
# Before (broken)
sources:
  - secret:
      name: {{ include "slurm.auth.secretName" $ }}

# After (fixed)  
sources:
  - configMap:
      name: {{ include "slurm.configMapName" $ }}
  - secret:
      name: {{ include "slurm.auth.secretName" $ }}
```

# Before fix:
```
[2025-05-23T14:46:50.178] warning: Can not stat gres.conf file (/var/spool/slurmd/conf-cache/gres.conf), using slurm.conf data
[2025-05-23T14:46:50.178] warning: Ignoring file-less GPU gpu:(null)
```

# After fix:
```
[2025-05-26T09:09:56.993] gres/gpu: init: loaded
[2025-05-26T09:09:56.993] gpu/nvidia: _set_name_and_file: Name=gpu Type=nvidia File=/dev/nvidia0

```
## Solution
Added the configmap to the projected volume sources to match the pattern used in the controller template (`helm/slurm/templates/controller/controller-statefulset.yaml` lines 103-108).

## How it was tested?
- GPU nodes now properly detect GPUs with `AutoDetect=nvidia` in `gres.conf`
- `sinfo` shows GPU nodes as `idle` instead of `inval`

### DIFF
--- a/helm/slurm/templates/compute/compute-nodeset.yaml
+++ b/helm/slurm/templates/compute/compute-nodeset.yaml
@@ -170,6 +170,8 @@ spec:
           projected:
             defaultMode: 0600
             sources:
+              - configMap:
+                  name: {{ include "slurm.configMapName" $ }}
               - secret:
                   name: {{ include "slurm.auth.secretName" $ }}
         - name: slurm-spool


SkylerMalinowski:
The proposed fix is not satisfactory. While it does fix the perceived issue, it introduces a regression.

Configless makes the slurmctld pod the source of truth for the slurmd, whom fetch their config files on init and reconfigure. By directly mounting the `slurm-config` volume on the slurmd pods, you are making their config files static (which is the regression).

When the slurm-config map

SkylerMalinowski:
Thank you for confirming pod recreation behavior (or lack thereof) when your configmap changes.

This indicates that commit 7436da26de01381b8b0364d01e5cc095bf41d84e introduced a reconfiguration regression, which you are experiencing right now.

Thanks for bringing this to our attention.

faganihajizada:
Happy to help @SkylerMalinowski. I am closing this PR.

SkylerMalinowski:
See commit fb21e7d8cfc809c907ac4a8786633f793959a35e for regression fix.

PR #15, opened 2025-06-05
Add per-nodeset authcred image override support

faganihajizada:
Adds support for configuring authcred (sackd) settings at the individual NodeSet level, allowing different nodesets to use different authcred images or resource configurations.

### Example Usage
```yaml
nodesets:
  - name: gpu
    authcred:
      imagePullPolicy: Always
      image:
        repository: ghcr.io/slinkyproject/sackd
        tag: 24.11-ubuntu24.04-custom
```

- Enables different authcred configurations per nodeset type (e.g., GPU vs CPU nodes)
- Maintains backward compatibility with existing global authcred configuration
- Provides flexibility for heterogeneous cluster deployments

### DIFF
--- a/helm/slurm/templates/compute/compute-nodeset.yaml
+++ b/helm/slurm/templates/compute/compute-nodeset.yaml
@@ -8,6 +8,16 @@ SPDX-License-Identifier: Apache-2.0
 {{- $repository := $nodeset.image.repository | default (include "slurm.compute.image.repository" $) }}
 {{- $tag := $nodeset.image.tag | default (include "slurm.compute.image.tag" $) }}
 {{- $imageRef := printf "%s:%s" $repository $tag }}
+{{- $authcredRepository := "" }}
+{{- $authcredTag := "" }}
+{{- if $nodeset.authcred }}
+{{- $authcredRepository = $nodeset.authcred.image.repository | default (include "slurm.authcred.image.repository" $) }}
+{{- $authcredTag = $nodeset.authcred.image.tag | default (include "slurm.authcred.image.tag" $) }}
+{{- else }}
+{{- $authcredRepository = include "slurm.authcred.image.repository" $ }}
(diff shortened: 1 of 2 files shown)


SkylerMalinowski:
Right now we use authcred (sackd) as a script runner side-car for the other daemons. Your request is not unreasonable but I do not necessarily want to enhance the hack we have. The authcred container should not require many resources so either a hard code or setting a small limit should be sufficient.

Perhaps you have a unique use case that we did not consider. Could you please elaborate on what exactly your use case is? Specifically, what are you doing that NodeSets would need or want to use a different authcred image?

faganihajizada:
> Right now we use authcred (sackd) as a script runner side-car for the other daemons. Your request is not unreasonable but I do not necessarily want to enhance the hack we have. The authcred container should not require many resources so either a hard code or setting a small limit should be sufficient.
> 
> Perhaps you have a unique use case that we did not consider. Could you please elaborate on what exactly your use case is? Specifically, what are you doing that NodeSets would need or want to use a different authcred image?

Thanks for detailed info @SkylerMalinowski 

In my case I have mixed CPU and GPU nodes in the cluster. Since the images provided for Slinky is not multi-arch, I had to build ARM images and schedule my GPU nodesets with these images. In that specific case, I need to be able to set `authcred` image for my nodeset as well otherwise it tries the default image which is for AMD

faganihajizada:
@SkylerMalinowski another solution might be to update containers repo and build multi-arch images for Slinky. Please let me know if that is preferred and happy to contribute.

I think there still might be another use cases for setting specific image for authcred

SkylerMalinowski:
Multi-arch container images would be preferred. It is on our radar as a priority, but contributions are certainly welcome. Thanks.

> I think there still might be another use cases for setting specific image for authcred

What would that be? I don't want to encourage customization of the authcred image given how we use it.

faganihajizada:
@SkylerMalinowski please let me know what do you think. thanks!

PR #16, opened 2025-06-09
Set non-zero TTL for token-create job

sd109:
When job TTL is zero the `helm install --wait --wait-for-jobs` command fails because the job is cleaned up by Kubernetes before Helm has a chance to see the job in a completed state which leads to the following error:
```
Error: INSTALLATION FAILED: jobs.batch "test-cluster-token-create" not found
```
Setting a short but non-zero TTL fixes the issue.

### DIFF
--- a/helm/slurm/templates/cluster/token-job.yaml
+++ b/helm/slurm/templates/cluster/token-job.yaml
@@ -11,7 +11,7 @@ metadata:
   labels:
     {{- include "slurm.token.labels" . | nindent 4 }}
 spec:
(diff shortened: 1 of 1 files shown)


sd109:
cc @SkylerMalinowski

SkylerMalinowski:
Thanks for the PR. I will look to land this fix. Thanks.

SkylerMalinowski:
https://github.com/SlinkyProject/slurm-operator/commit/44b79099e7caddb0a7f1c9a078e6ca7747b31cfc

PR #22, opened 2025-07-15
Refactor slurm configuration management

faganihajizada:
## Overview

The current architecture creates **duplicate configuration entries** when users try to override hardcoded values in templates. For example:

**Before (problematic):**
- Template hardcoded: `GresTypes=gpu` 
- User's extraSlurmConf: `GresTypes: gpu,mps,fpga`
- **Result**: Both entries appear in final `slurm.conf`, causing unpredictable behavior

Instead of specifying extra configuration options as a `str:str map`, passing in an extra.conf file via configmaps and include it in the main config is clear approach.

## Need help

I moved some values from templates to user-configurable defaults:
- `GresTypes=gpu`
- `ReturnToService=2` 
- `MaxNodeCount=1024`
- `LogTimeFormat=iso8601_ms,format_stderr`
- `SelectTypeParameters=CR_Core_Memory`
- `CommunicationParameters=block_null_hash` 
- `LaunchParameters=enable_nss_slurm,use_interactive_step,ulimit_pam_adopt`
- `JobAcctGatherFrequency=task=30`
- `AccountingStorageTRES=gres/gpu`
- `LogFile=/dev/null` (slurmdbd)

I see there are many hardcoded values and I need your help to decide which one should stay hardcoded in helm configmap tempalates. We should minimize the hardcoded ones.


SkylerMalinowski:
Of the items you listed, only the following are strictly required.

```ini
# Needed because of AuthInfo=use_client_ids and lack of sssd in
# slurmctl and slurmd pods.
LaunchParameters=enable_nss_slurm

# Must be > 0 for dynamic nodes to work in this case.
# 1024 was arbitrarily chosen to be reasonably high but not too high. 
MaxNodeCount=1024

# SelectTypeParameters must be set, otherwise fatal.
# CR_Core_Memory is the most common.
SelectTypeParameters=CR_Core_Memory
```

All the other values are set with sanity or better defaults than what Slurm does.

```ini
# format_stderr makes it closest to the logfile logging.
LogTimeFormat=iso8601_ms,format_stderr

# A good default for cloudy environments.
ReturnToService=2

# Adds additional security but may not be needed.
CommunicationParameters=block_null_hash

# GPUs are are a well known and used.
GresTypes=gpu
AccountingStorageTRES=gres/gpu
```

faganihajizada:
@SkylerMalinowski for the idea to have ability for both, do you have any implementation specific suggestion since you mentioned bundling the proposed *_extra.conf under configFiles will not be convenient?

SkylerMalinowski:
Offering both methods, string and/or map, seems best. Similar patterns are used by other charts but the implement varies. One overrides the other vs merge both are common implementations. Not sure which is better, but I see the override being more common. 

If we did refactor out the optional config items into the default extraConf, now we have to choose to express as a string or map. Which takes presence, string or map?

> @SkylerMalinowski for the idea to have ability for both, do you have any implementation specific suggestion since you mentioned bundling the proposed *_extra.conf under configFiles will not be convenient?

Each extraConf was moved under their respective object (e.g. controller.extraConf, accounting.extraConf), coinciding with their CRDs. Conf generation was internalized into the operator.

Technically how you rigged up configFiles with `*_extra.conf` can still work with the new stuff. However `slurmdbd_extra.conf` would be the odd config out. configFiles is a map containing config files for the controller/cluster (excluding accounting). My initial thought is: don't conflate configFiles in that way.

It is not straightforward (or at least clean) how to support both string and map expressions for config files. Typically a similar values field is added to handle the other type (e.g. extraConf (string), extraConfMap (map)). Following this methodology, we would need configFiles (map of string) and configFilesMap (map of map). And possibly do the same for accounting, especially if we did the `*_extra.conf` idea. Doable but doesn't feel ideal.

I will certainly need to think about this more. Everyone's input has been helpful. Thanks.

faganihajizada:
Great! thank you @SkylerMalinowski

PR #23, opened 2025-07-16
feat: add topology.yaml to config files

dmitsh:
Adding topology.yaml to the list of SLURM config files.


alanmutsch:
Thanks for the PR, I'll work on getting this merged.

PR #39, opened 2025-08-27
Remove automatic DaemonSet tolerations from NodeSet pods

faganihajizada:
## Summary

Removes automatic DaemonSet tolerations from NodeSet pods to prevent them from scheduling on cordoned nodes during maintenance.

**Problem**: NodeSet pods currently inherit DaemonSet tolerations (including `node.kubernetes.io/unschedulable`), allowing them to bypass Kubernetes node cordoning. This defeats the purpose of cordoning nodes for maintenance and can lead to workloads running on unstable nodes.

**Root Cause**: The `daemonutil.AddOrUpdateDaemonPodTolerations()` call was retained from the original StatefulSet+DaemonSet implementation.

**Solution**: Remove the automatic DaemonSet tolerations. Users who need specific tolerations can configure them explicitly via the NodeSet spec's `tolerations` field.

## Breaking Changes

N/A

The change only removes _automatic_ tolerations that were overriding user intent and standard Kubernetes scheduling behavior.

## Testing Notes

**Tested:**
- All existing tests pass
- NodeSet pod creation works correctly
- User-specified tolerations are preserved

### DIFF
--- a/internal/controller/nodeset/utils/utils.go
+++ b/internal/controller/nodeset/utils/utils.go
@@ -13,7 +13,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	k8scontroller "k8s.io/kubernetes/pkg/controller"
-	daemonutil "k8s.io/kubernetes/pkg/controller/daemon/util"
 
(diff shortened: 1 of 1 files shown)


faganihajizada:
cc @vivian-hafener

SkylerMalinowski:
LGTM. @vivian-hafener , go ahead with getting this merged in.

PR #58, opened 2025-10-01
Add support for different service types in loginset

faganihajizada:
## Summary

Add support for different service types (NodePort, LoadBalancer) to loginset while maintaining default ClusterIP behavior. This enables use cases where external access to login nodes is needed via NodePort.

## Breaking Changes

N/A

## Testing Notes

Tested internally with values below and works as expected

```
loginsets:
  slinky:
    enabled: true
    service:
      type: NodePort
      port: 22
      nodePort: 32222
```

## Additional Context

Alternative option might be to update **ServiceSpec** in go code


faganihajizada:
Good morning, @vivian-hafener 

Yes it does. thank you! I missed that, my bad

PR #132, opened 2026-02-27
fix: nil pointer dereference when RollingUpdate is nil

antoinerrr:
## Summary

Fix a nil pointer dereference panic in `splitUpdatePods` when `UpdateStrategy.Type` is `RollingUpdate` but `UpdateStrategy.RollingUpdate` is not set.

## Breaking Changes

None

## Testing Notes

Pretty self explanatory, on main, if you create a nodeset and try to scale it when the update strategy is RollingUpdate but you do not set UpdateStrategy.RollingUpdate.MaxUnavailable variable, it will panic. This PR add a default 

## Additional Context

None

## Fix

Add a nil guard before accessing RollingUpdate.MaxUnavailable. When RollingUpdate is nil, a nil *IntOrString is passed to GetScaledValueFromIntOrPercent, which already handles it by returning the default value (1).

### DIFF
--- a/internal/controller/nodeset/nodeset_sync.go
+++ b/internal/controller/nodeset/nodeset_sync.go
@@ -14,6 +14,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	k8slabels "k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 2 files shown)


SkylerMalinowski:
We have solved this with a different implementation. Please see 7886d0ec7af8f46c47cc4d3614dd024c1f061ad0 for details.

PR #141, opened 2026-03-17
fix(nodeset): use bare node name as DaemonSet pod hostname

giuliocalzo:
## Summary
- DaemonSet-mode NodeSet pods now use the Kubernetes node name directly as the pod hostname instead of prepending a prefix.
- This avoids duplicated names (e.g. `cpu-cpu-zzzz-xxxx`) when the node name is `cpu-zzzz-xxxx`.

## Breaking Changes

none

## Testing Notes

- [x] Updated `TestGetDaemonSetPodHostname` expectations for no-prefix cases
- [x] Verified custom hostname prefix tests still pass unchanged
- [x] Deploy a DaemonSet-mode NodeSet and confirm pods use bare node names as hostnames


giuliocalzo:
merged in https://github.com/SlinkyProject/slurm-operator/commit/1db558d903e9e41b2807afc287ae24bc18370e5c

PR #148, opened 2026-03-19
Remove dead `condChanged` variable

faganihajizada:
## Summary

Remove the dead `condChanged` variable in `updateNodeSetPodConditions`. The variable was assigned but never read after the loop — the `!condChanged` guard only prevented redundant re-assignment of the bool itself, with no downstream effect. `UpdatePodCondition` is still called for its side effect of mutating `toUpdate.Status`.

## Breaking Changes

N/A

## Testing Notes

Existing unit tests pass (`TestNodeSetReconciler_updateNodeSetPodConditions`)


vivian-hafener:
This was squashed into https://github.com/SlinkyProject/slurm-operator/commit/b5fd0caa6d372af20e252afb76000a32148cded9 and merged.

PR #158, opened 2026-03-25
feat: add SlurmQos CRD for declarative QOS management

faganihajizada:
## Summary

Add declarative Slurm QOS management via a new `SlurmQos` CRD with controller and webhook. QOS policies are reconciled against the Slurm REST API (slurmdb endpoints), writing directly to the accounting database: no `scontrol reconfigure` required.

This supports the strategic shift from partition-centric to QOS-centric scheduling: fewer partitions, fine-grained preemption hierarchies, live policy updates, and lower scheduler load.

**New CRD: `SlurmQos`**
- `controllerRef` scopes to a Controller CR (requires accounting enabled)
- `description` for human-readable QOS description
- `config` as `runtime.RawExtension`
- `metadata.name` maps directly to the Slurm QOS name

**Controller** (`internal/controller/slurmqos/`):
- Reconciles SlurmQos CRs via ClientMap -> slurmrestd -> slurmdbd
- Finalizer-based deletion cleanup
- Status with `ready`, `id` (Slurm-assigned), and `Synced` condition

**Webhook** (`internal/webhook/slurmqos_webhook.go`):
- Validates controllerRef, QOS name format, JSON config, accounting requirement
- Immutable controllerRef on update
- Warns on "normal" QOS (Slurm creates one by default)

**Helm chart**:
- `qos` map in values.yaml (gated by `accounting.enabled`)
- Template renders SlurmQos CRs per entry
- 7 unit tests

**Depends on:** https://github.com/SlinkyProject/slurm-client/pull/2

## Breaking Changes

N/A

## Testing Notes

- `helm unittest helm/slurm/ --file tests/qos_test.yaml`
- Webhook tests: `go test ./internal/webhook/ -run SlurmQos
- E2E verified on kind cluster with accounting enabled:
  - Create: QOS appears in Slurm with correct priority, flags, preempt mode
  - Update: `kubectl patch` reflected in Slurm within seconds
  - Delete: finalizer removes QOS from Slurm, then CR is garbage collected
  - Accounting guard: `Synced=False` with clear message when accountingRef is nil
- `make generate && make manifests`

## Additional Context

- The `config` field uses `runtime.RawExtension` which is new to this operator's CRDs.
  Justified because QOS config maps to arbitrary JSON for the REST API (unlike partition config which is a flat key=value string). This is the standard K8s approach for embedding opaque JSON in CRDs.
- QOS names are bare (not prefixed with chart fullname) because `metadata.name` must match the exact Slurm QOS name. The `controllerRef` provides cluster scoping.
- Account-QOS associations (`accounts={"nvidia": "denied"}`) are out of scope: they require a separate `SlurmAssociation` CRD as a follow-up.
- Flag values must use V0044 API enum format (e.g. `DENY_LIMIT`, not `DenyOnLimit`).

### Known Limitations
- **Account-QOS associations are out of scope.** QOS definitions that require account restrictions (e.g. `accounts={"account_name": "denied"}`) cannot be managed via the `V0044Qos` REST API. These require `sacctmgr` associations. Follow-up: a `SlurmAssociation` CRD using the same architecture pattern. The `slurm-client` already has generated OpenAPI types for associations. I will create separate PR.
- **Dynamic TRES limits are set manually per cluster.** For example, the `normal` QOS needs `max_tres_pu.nodes = 0.5 * cluster_node_count`. Users must compute and set the absolute value at deploy time. The Helm values comments document the formula.


SkylerMalinowski:
We do not want to codify slurmdbd user/account/assoc management in this way. We are working with Slurm maintainers to help drive towards declarative resource management in slurmdbd. Once that lands in Slurm, we can revisit this topic in Slinky.

PR #173, opened 2026-04-14
feat: prune defunct slurm nodes for daemonset nodesets

kaynetu:
## Summary

Add defunct Slurm node pruning for NodeSets running in `DaemonSet` mode.

This addresses the case where a Kubernetes worker node disappears before the worker pod's `preStop` hook can clean up its Slurm node record. In that failure mode, Slurm can retain a stale node entry indefinitely, which creates noise in monitoring and operational views.

This change prunes a Slurm node only when all of the following are true:
- the NodeSet is using `DaemonSet` scaling mode
- the Slurm node is `DOWN+NOT_RESPONDING`
- the Slurm node contains valid `PodInfo`
- the referenced Kubernetes pod no longer exists
- the referenced Kubernetes node no longer exists

The implementation is intentionally limited to `DaemonSet` mode because in that mode the Slurm node identity follows the Kubernetes node identity. In `StatefulSet` mode, Slurm node identity is stable with the pod name, so replacement pods preserve the same Slurm node record instead of leaving behind a defunct node.

## Breaking Changes

None intended.

The only behavioral change is that the operator may now delete stale Slurm node records for `DaemonSet` NodeSets when they are clearly defunct by the criteria above.

## Testing Notes

Unit tests:
- added coverage for defunct-node discovery and deletion in `slurmcontrol`
- added coverage for `syncDefunctNodes` in `DaemonSet` mode
- added coverage for the skip path in `StatefulSet` mode

Manual validation on a local `kind` cluster:
1. Run the worker NodeSet in `DaemonSet` mode.
2. Remove a worker node at the infrastructure level.
3. Delete the Kubernetes `Node` object.
4. Wait for Slurm to transition the corresponding node to `DOWN+NOT_RESPONDING`.
5. Verify the operator logs show defunct-node pruning and the stale Slurm node is removed.

Operator log during validation:
```json
{"level":"info","ts":"2026-04-14T02:26:35Z","msg":"Started syncing NodeSet","controller":"nodeset-controller","controllerGroup":"slinky.slurm.net","controllerKind":"NodeSet","NodeSet":{"name":"slurm-worker-slinky","namespace":"slurm"},"namespace":"slurm","name":"slurm-worker-slinky","reconcileID":"554f3880-5334-4de8-9240-5beb16be4a78","request":{"name":"slurm-worker-slinky","namespace":"slurm"}}
{"level":"info","ts":"2026-04-14T02:26:36Z","msg":"Deleting defunct Slurm node without a corresponding Kubernetes Pod","controller":"nodeset-controller","controllerGroup":"slinky.slurm.net","controllerKind":"NodeSet","NodeSet":{"name":"slurm-worker-slinky","namespace":"slurm"},"namespace":"slurm","name":"slurm-worker-slinky","reconcileID":"554f3880-5334-4de8-9240-5beb16be4a78","node":"kind-worker2","pod":{"name":"slurm-worker-slinky-fqb2j","namespace":"slurm"},"kubeNode":"kind-worker2"}
{"level":"info","ts":"2026-04-14T02:26:48Z","msg":"Finished syncing NodeSet","controller":"nodeset-controller","controllerGroup":"slinky.slurm.net","controllerKind":"NodeSet","NodeSet":{"name":"slurm-worker-slinky","namespace":"slurm"},"namespace":"slurm","name":"slurm-worker-slinky","reconcileID":"554f3880-5334-4de8-9240-5beb16be4a78","duration":13.101964382,"result":{"Requeue":false,"RequeueAfter":30000000000,"Priority":null}}
```

## Additional Context

This reintroduces defunct-node pruning with stricter safety checks than the [earlier removed implementation](https://github.com/SlinkyProject/slurm-operator/commit/aa44fcc8).

The pruning path is restricted to `DaemonSet` mode because the additional Kubernetes node existence check is meaningful only when the Slurm node is derived from the Kubernetes node. In `StatefulSet` mode, requiring the Kubernetes node to be gone would be too strict and could block cleanup incorrectly.

This implementation was validated against the target failure mode where node loss prevents `preStop` from running and leaves a stale Slurm node record behind.


internal/controller/nodeset/nodeset_sync.go
```diff
@@ -532,6 +538,77 @@ func (r *NodeSetReconciler) syncNodeTaint(
...
+
+	// This pruning logic only applies to DaemonSet mode because the Slurm node
+	// identity is derived from the Kubernetes node identity. If a Kubernetes node
+	// disappears, its Slurm node record can become defunct until cleaned up.
+	// In StatefulSet mode, the Slurm node identity is stable with the pod, so a
+	// replacement pod comes back with the same Slurm node name and preserves that
+	// history instead of leaving behind a ghost/defunct node record.
+	if nodeset.Spec.ScalingMode != slinkyv1beta1.ScalingModeDaemonset {
+		return nil
+	}
```
SkylerMalinowski, line 548: I think it better to make this work on the entire NodeSet regardless of scaling mode.

In both scaling type cases, if the default preStop fails to succeed, node records will linger until a pod with the same hostname registers (restoring the node record as active) or the admin manually deletes the node record. In StatefulSet, scaling up can restore the node record, but there is no guarantee the replicas can or will be scaled up and when. In DaemonSet, a removed Kube node may never come back (with the same name), hence the ghost node may never be restored naturally.

internal/controller/nodeset/nodeset_sync.go
```diff
@@ -532,6 +538,77 @@ func (r *NodeSetReconciler) syncNodeTaint(
...
+			logger.Info("Skipping defunct Slurm node deletion because the Kubernetes node still exists",
+				"node", defunctNode.Name, "pod", podKey, "kubeNode", kubeNodeKey.Name)
+			return nil
+		} else if !apierrors.IsNotFound(err) {
+			return err
+		}
+
+		// Both the pod and the Kubernetes node are gone — prune the Slurm node.
+		logger.Info("Deleting defunct Slurm node without a corresponding Kubernetes Pod",
+			"node", defunctNode.Name, "pod", podKey, "kubeNode", kubeNodeKey.Name)
```
SkylerMalinowski: Logging should use the Kubernetes types, but Slurm types should be prefixed with "slurm" to differentiate.

```go
			"slurmNode", defunctNode.Name, "pod", podKey, "node", kubeNodeKey.Name)
```

internal/controller/nodeset/slurmcontrol/slurmcontrol.go
```diff
@@ -625,6 +634,81 @@ func (r *realSlurmControl) GetNodesForPods(ctx context.Context, nodeset *slinkyv
...
+	}
+
+	slurmNode := &slurmtypes.V0044Node{}
+	key := slurmobject.ObjectKey(nodeName)
+	if err := slurmClient.Get(ctx, key, slurmNode); err != nil {
+		if tolerateError(err) {
+			return nil
+		}
+		return err
+	}
```
SkylerMalinowski: An optimization here is to not do the `Get()` before the `Delete()`. Prepare the `V0044Node{}` such that its Name is set to the nodeName, then call `Delete()` with the object. This works similar to the Kubernetes client.

SkylerMalinowski (commented):
Looks pretty good. There are a few little things to address.

The only design issue I have is limiting this Slurm node record pruning to only DaemonSet scaled NodeSets. As pointed out in the review comment, both StatefulSet and DaemonSet scaled pods can suffer from ghost/defunct node records.

internal/controller/nodeset/nodeset_sync.go
kaynetu: My thoughts are: 
* Only in daemonset mode, a slurm node will come back with a new name. The old node lingers and messes up the metric
* In Statefulset mode, though it could happen that preStop step doesn't run, later, when an eligible k8s node is up (whether it's on the new one or still on the old one), it will create the same ordinal-based name (slurm-worker-2), it won't become a ghost node. 

Please let me know what you think on this

kaynetu: Had a great discussion with Skyler offline. There are 3 requirements: 
1. If a Slurm node is gone forever, we don't leave the DOWN node messing up with metrics
2. Adding this node record deletion will catch cases where preStop fails to delete the node.
3. Keep Slurm node history if we know the issue is transient: meaning. the node is not gone forever.

This matters for daemonset mode as we use k8s hostname as Slurm node name, a new node will have new hostname and the old node is gone forever. If the k8s node is not replaced, we don't delete Slurm node and history is kept. 

We've decided to keep the new step restrictive to only daemonset mode first. Let StatefulSet mode rely on existing preStop hook.  If we find cases later, we can open this up. It's a two-way door decision

SkylerMalinowski: To add context: this node record pruning logic makes the most sense when preStop does not attempt to delete the Slurm node. Hence all Slurm node records linger, and the new sync step handles an edge case where a Kube node is deleted which previously had a NodeSet pod (daemonSet mode) on it, so the record should be cleaned up.

yelsayd:
I think this is great to have. But can we condition this behind a spec config? It can be surprising to slurm admins when nodes disappear completely from the cluster, the `DOWN` + `UNRESPONSIVE` states exist for this reason (e.g. machine is down for maintenance and is expected to be back in O(days)).

I think this is coming from an assumption that the underlying host can disconnect and come back with a different slurm node name, which is true only in some configurations, but not always true.

SkylerMalinowski:
> I think this is great to have. But can we condition this behind a spec config? It can be surprising to slurm admins when nodes disappear completely from the cluster, the `DOWN` + `UNRESPONSIVE` states exist for this reason (e.g. machine is down for maintenance and is expected to be back in O(days)).

Less is more. I do not want the Slurm Operator be be a hydra of configuration options. Remember, the operator services all Slurm clusters deployed by it, not just yours. Therefore adding an option to gate this logic should be a last resort, as far as I'm concerned.

> I think this is coming from an assumption that the underlying host can disconnect and come back with a different slurm node name, which is true only in some configurations, but not always true.

This is true. The more Cloud Native oriented defaults need to be reconsidered (and perhaps changed) now that DaemonSet scaling style was added.

kaynetu:
In the latest two commits, I have
* Use nodeset config to opt-in/out the node pruning behavior. Defaults to false
* Dealt with the [node annotation overriding Slurm node name](https://github.com/SlinkyProject/slurm-operator/commit/fbf5f43801515233430aabbb8f9236331b239cd8)
* Consolidated all new syncDefunctNodes tests into one 
* Removed `scontrol delete` from preStop hook

alanmutsch:
Merged 99bdaa418a625ee8d8858b0f856e0351187b51b1

PR #175, opened 2026-04-15
test(helm): extend NodeSet tests for DRA-style claims and assertions

giuliocalzo:
## Summary

Extends `helm/slurm/tests/nodeset_test.yaml` to cover Dynamic Resource Allocation-style pod configuration and to tighten a few assertions.

## Changes

- **Resource claims:** New test for `podSpec.resourceClaims` (asserted with `contains`) and `podSpec.resources.claims` (asserted with `equal` on `spec.template.spec.resources`).
- **Resources only:** Renamed/split the prior limits/requests case into `should set pod template resources` without mixing in claim fields.
- **Affinity:** Assert the full `spec.template.spec.affinity` object instead of a single nested leaf path.
- **Topology spread:** Use `contains` on `spec.template.spec.topologySpreadConstraints` with the full constraint object.
- **Formatting:** Blank lines between test cases for readability.

## Verification

```bash
helm unittest helm/slurm -f 'tests/nodeset_test.yaml'
```


klairecodes (approved):

vivian-hafener (changes requested):
I've requested changes in order to block the merge of this PR. We handle PR merges internally, instead of using GitHub, so we will handle the merge internally. Will close once merged internally per usual, just don't want to risk an accidental merge to `main` here as it is annoying to resolve.

klairecodes (changes requested):
Blocking merge.

PR #177, opened 2026-04-16
feat(restapi): configurable PodDisruptionBudget for HA

giuliocalzo:
## Summary

This change adds optional **PodDisruptionBudget** support for the RestApi workload when running with more than one replica, so voluntary disruptions respect standard Kubernetes PDB semantics. It also tightens validation and safe reconcile behavior around edge-case PDB specs.

## What changed

- **API**: `RestApiSpec.podDisruptionBudget` with an `enabled` flag and an **inline** `policy/v1.PodDisruptionBudgetSpec` (the operator overwrites the label **selector** to match the RestApi Deployment).
- **Controller**: Creates, updates, or deletes the PDB from replica count and `enabled`; nil `enabled` keeps PDB on for HA (replicas > 1). If `minAvailable` would require **every** replica (integer equal to replica count, or `100%`), the controller **does not** apply a PDB, emits a **Warning** event (`PodDisruptionBudgetSkipped`), removes any existing PDB, and continues reconcile without failing the sync step.
- **Builder**: Copies the embedded spec into the PDB object; if both `minAvailable` and `maxUnavailable` are unset, defaults `minAvailable` to `1`.
- **Validating webhook**: When **replicas ≤ 1** and `podDisruptionBudget` is **enabled** (default), PDB field validation is **skipped** because no PDB is reconciled. When **replicas ≥ 2** (or PDB is disabled with a single replica), **both** `minAvailable` and `maxUnavailable` set is rejected as invalid.
- **Helm / CRDs**: Values, RestApi CR template, CRD schema, README, and chart snapshot tests.

## Testing

- `go test ./internal/controller/restapi/... ./internal/builder/restapibuilder/... ./internal/webhook/...`
- Envtests in `restapi_controller_test.go` for PDB present/absent cases.
- Unit tests for PDB desired / min-demands-all-replicas helpers and RestApi webhook PDB validation.


vivian-hafener:
Good afternoon Giulio.

It would be preferable if this implementation did not embed the PDB spec into the RestAPI CRD. That increases support burden over time and couples us more tightly with versioning of the PDB spec. While elsewhere in the codebase we avoid direct struct embedding using "wrappers", I'd prefer that in this case you try to refactor this such that the PDB for the RestAPI is deployed & configured via the Helm charting.

Best,
Vivian Hafener

giuliocalzo:
makes sense... to be fair having the pdb only in helm is even easier! let's close this PR I will make a new one

giuliocalzo:
Superseded by #201, which implements the RestAPI PodDisruptionBudget purely at the Helm chart level (no CRD/operator changes), per the review feedback.

PR #180, opened 2026-04-21
feat(loginset): support OnDelete strategy backed by a StatefulSet

antoinerrr:
## Summary

Extends `LoginSet.Spec.Strategy.Type` to accept `OnDelete` in addition to the standard Deployment values (`RollingUpdate`, `Recreate`). When set, the LoginSet renders as a StatefulSet whose `updateStrategy.type` is `OnDelete`, so pod template changes (e.g. an image update) only roll when an operator deletes the Pod.

This is intended for login pods with active interactive SSH sessions, where the existing Deployment-backed strategies always tear down running pods on spec change.

The default behavior (Deployment + `RollingUpdate`) is unchanged, and existing users of `Strategy` keep working without modification.

## Checklist

## Breaking Changes

None. The change is purely additive: existing `Strategy: {type: RollingUpdate, ...}` and `Strategy: {type: Recreate}` configurations render the same Deployment as before.

### Note on the design choice

I considered two designs:

1. **Additive (this PR)**: keep `Spec.Strategy` (`appsv1.DeploymentStrategy`) and accept a third value `OnDelete` for `Type` that selects a StatefulSet workload.
2. **Destructive**: replace `Spec.Strategy` with a new `LoginSetUpdateStrategy` type (a discriminator mirroring `NodeSetUpdateStrategy`) that explicitly models the workload kind.

I went with (1) to avoid removing the `Strategy` field that was just added in #b416a90 and to keep the upgrade path trivial. The cost is that `Strategy.Type` now accepts a value (`OnDelete`) that is not native to `appsv1.DeploymentStrategyType`, which is a slight semantic stretch documented on the field. Happy to switch to (2) if maintainers prefer the cleaner type at the cost of churning the field.

## Testing Notes
- New tests:
  - `loginbuilder.TestBuilder_BuildLoginStatefulSet` and `TestBuilder_BuildLoginWorkload` cover the dispatch and StatefulSet shape.
  - `loginset.TestControllers` gains specs for "renders a StatefulSet under OnDelete" and "garbage-collects the stale Deployment when switching to OnDelete".
  - `webhook.TestAPIs` gains 6 strategy-validation cases (valid/invalid `Type`, `RollingUpdate` combined with non-`RollingUpdate` types).

To exercise manually:

```yaml
apiVersion: slinky.slurm.net/v1beta1
kind: LoginSet
spec:
  strategy:
    type: OnDelete
  # ...
```

Update the `login.image` and observe that the existing pod is not rolled until you `kubectl delete pod <login-pod>`.

## Additional Context


vivian-hafener:
Good morning @antoinerrr,

Could you provide more information on why you want LoginSets to be a StatefulSet? Is there some use-case that StatefulSets are better for, or a way that they improve the user experience of LoginSets?

We likely won't move forward with dynamically switching the backing type for LoginSets between Deployments and StatefulSets based on the contents of the `UpdateStrategy` variable.  However, if there's a use-case that would benefit from switching LoginSets to StatefulSets across the board, I may be inclined to implement such a solution myself after some discussion with the rest of the team.

Best,
Vivian Hafener

antoinerrr:
@vivian-hafener 
Hi!
We use personal login pod, having no way to let a user choose the best moment for his environment to be deleted and recreated (ex: image change) is a real issue for us. We are use to this behavior with Sunk that actually use statefulset and not having this flexibility is a huge pain point for us. Maybe there is an other way around this ?
I would be more than happy to help if you agree to go in the full stateful direction

vivian-hafener:
Good afternoon @antoinerrr,

In the long run we are considering moving towards per-user ephemeral login pods. That would probably be the best solution for your environment. However, we have not decided on an implementation or design for that, nor have we decided what release that will be targeted for. Due to the scope of that change, we intend to handle it entirely in-house. 

Our current LoginSet CR is intended to be used in a manner that is analogous to the traditional HPC model, wherein there's a smaller number of multi-tenant Login nodes. While I understand that that model may not be ideal for your use-case, I mostly provide this context to explain the design behind our current model.

I will assess how much effort it would take to switch our implementation of LoginSets to StatefulSets from Deployments across the board, and will check in with some other Slinky users to ensure that this change would be beneficial in their environments as well.

Best,
Vivian Hafener

antoinerrr:
Hi @vivian-hafener . Thanks for your feedback

antoinerrr:
@vivian-hafener would you be open to me opening a draft PR implementing the full switch to kickstart the discussion ? (Sunk / Soperator both use StatefulSet btw)

vivian-hafener:
Good morning @antoinerrr,

I would recommend that your users make use of the `corev1.Deployment.Spec.Paused` field in order to prevent updates to their LoginSets until they are ready: https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#paused. It sounds like this provides all of the capability that you are looking for out of a transition to StatefulSets, specifically from the `updateStrategy.OnDelete` field.

Upon further discussion we have decided not to move forward with this change. `Deployment.Spec.Paused` provides a reasonable workaround for your use-case, in that it behaves similarly to `updateStrategy.OnDelete`. In most use-cases, LoginSet pods shouldn't require the stable storage or ordered startup provided by the StatefulSet type, so switching types would increase complexity with little benefit over the `Deployment.Spec.Paused` solution.

Slurm-operator has a distinct and unique advantage in this space in that it is developed in close collaboration with the teams that wrote and developed Slurm. As such, we sometimes make different choices from other tools for Slurm-on-Kubernetes, based on our knowledge of both the architecture of Slurm and the design decisions behind that architecture.

Please feel free to reach out should you have any further questions on the matter.

Best,
Vivian Hafener

PR #181, opened 2026-04-23
fix(chart): dcgm.enabled=false now disables dcgm

faganihajizada:
## Summary

Restore the `vendor.dcgm.enabled` helper so `dcgm.enabled: false` actually disables the DCGM integration.

Since [6f6668d](https://github.com/SlinkyProject/slurm-operator/commit/6f6668db87db59d4cbcbaf41876b6294137c5fb6) (`refactor(chart): use raw bool, not redundant ternary`) the helper emits the raw bool value of `.Values.vendor.nvidia.dcgm.enabled`. Because `include` always returns a string, a disabled value renders as the literal string `"false"` — which is truthy in Go templates. Every `{{- if (include "vendor.dcgm.enabled" .) }}` gate therefore evaluates true regardless of configuration, and DCGM resources are injected even when users set `dcgm.enabled: false`:

- `prologScriptRefs` / `epilogScriptRefs` on the Controller CR gain the DCGM prolog/epilog configmap references.
- Every GPU-requesting NodeSet gets the `dcgm-job-mapping` hostPath volume and its volumeMount appended.

The "redundant" ternary was actually doing essential string coercion (`"true"` / `""`), so this PR reverts the helper to that form and adds negative-path coverage for every `include "vendor.dcgm.enabled"` call site so the regression can't come back silently.

Closes: user reports of DCGM volume mounts appearing despite `dcgm.enabled: false`.

## Checklist

- [x] Documentation is updated if user-visible behavior changes. (No user-facing docs change; values schema and defaults are unchanged.)

## Breaking Changes

N/A

## Testing Notes

Ran `helm unittest --strict ./helm/slurm` — **113/113 tests pass, 5/5 snapshots pass**.

Added 7 regression tests in `helm/slurm/tests/vendor_dcgm_test.yaml` covering every `include "vendor.dcgm.enabled"` consumer:

- prolog/epilog configmaps are not rendered when disabled
- controller `spec.prologScriptRefs` / `spec.epilogScriptRefs` are empty when disabled and populated when enabled
- nodeset `spec.template.spec.volumes` and `spec.slurmd.volumeMounts` do not contain `dcgm-job-mapping` when disabled
- nodeset correctly injects the volume + mount when enabled AND the nodeset requests `nvidia.com/gpu`
- nodeset does NOT inject the volume + mount when enabled but no GPUs are allocated (guards the `nodesetHasGPU` path)

Confirmed the new tests fail against the broken helper (4 assertions fail, reproducing the exact user-reported symptoms) and pass with the fix.

## Additional Context

The controller_test.yaml snapshot was captured after 6f6668d landed, so it had baked in the bug (DCGM script refs appearing in a manifest rendered with the default dcgm.enabled: false). Regenerating the snapshot as part of this PR removes those lines — which is itself a nice confirmation of the regression.

### DIFF
--- a/helm/slurm/templates/vendor/nvidia/dcgm/_helpers.tpl
+++ b/helm/slurm/templates/vendor/nvidia/dcgm/_helpers.tpl
@@ -8,7 +8,7 @@ SPDX-License-Identifier: Apache-2.0
 Check if DCGM integration is enabled
 */}}
 {{- define "vendor.dcgm.enabled" -}}
-{{- .Values.vendor.nvidia.dcgm.enabled -}}
+{{- .Values.vendor.nvidia.dcgm.enabled | ternary "true" "" -}}
 {{- end }}
 
 {{/*
--- a/helm/slurm/tests/__snapshot__/controller_test.yaml.snap
+++ b/helm/slurm/tests/__snapshot__/controller_test.yaml.snap
@@ -11,8 +11,7 @@ manifest should match snapshot:
       name: test-release-slurm
       namespace: test-namespace
     spec:
-      epilogScriptRefs:
-        - name: test-release-slurm-epilog-dcgm
+      epilogScriptRefs: null
       epilogSlurmctldScriptRefs: null
       extraConf: PartitionName=all Nodes=ALL Default=YES MaxTime=UNLIMITED State=UP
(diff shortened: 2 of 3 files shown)


SkylerMalinowski:
Merged: f1786e45

PR #182, opened 2026-04-24
feat(helm): expose topologySpreadConstraints on operator and webhook

giuliocalzo:
## Summary

The `slurm-operator` chart surfaced `affinity` and `tolerations` on both the operator and webhook Deployments, but not `topologySpreadConstraints`. With `replicas > 1` and leader election enabled, Kubernetes was free to schedule every replica onto the same node or zone, which undermines the HA story the chart otherwise enables.

The sibling `slurm` chart already plumbs `topologySpreadConstraints` through its component pod specs (controller, login, accounting, etc.), so this brings the `slurm-operator` chart to feature parity.

Changes:

- `helm/slurm-operator/values.yaml`: add opt-in `operator.topologySpreadConstraints` and `webhook.topologySpreadConstraints` keys (default `[]`) with inline commentary and an example zone-spread block.
- `helm/slurm-operator/templates/operator/deployment.yaml` and `helm/slurm-operator/templates/webhook/deployment.yaml`: render a `topologySpreadConstraints:` block via `{{- with ... }}` right after the existing `tolerations` block, so output is byte-identical when the value is left at its default.
- `helm/slurm-operator/tests/operator_deployment_test.yaml` and `helm/slurm-operator/tests/webhook_deployment_test.yaml`: add a positive rendering case and a `notExists` default-off case per suite.

## Checklist

- [x] Documentation is updated if user-visible behavior changes. (Values docs updated inline; no user-facing README changes needed since the new keys are opt-in and documented where all other scheduling hints live.)

## Breaking Changes

None. Both new keys default to `[]`, which keeps the rendered YAML byte-identical for existing installs. All pre-existing `helm-unittest` snapshot tests still match without updates.

## Testing Notes

```
helm lint helm/slurm-operator
helm unittest helm/slurm-operator
helm template helm/slurm-operator \
  --set operator.topologySpreadConstraints[0].maxSkew=1 \
  --set operator.topologySpreadConstraints[0].topologyKey=topology.kubernetes.io/zone \
  --set operator.topologySpreadConstraints[0].whenUnsatisfiable=ScheduleAnyway \
  --set operator.topologySpreadConstraints[0].labelSelector.matchLabels."app\.kubernetes\.io/name"=slurm-operator
```

All three commands succeed, snapshots are unchanged, and the `topologySpreadConstraints` block appears on the rendered Deployment only when the new value is set.

## Additional Context

The chart already exposes `nodeSelector`, `affinity`, and `tolerations` on the same Deployments, so `topologySpreadConstraints` was the only scheduling primitive missing. This is intentionally a minimal, additive change; no defaults are altered.

### DIFF
--- a/helm/slurm-operator/README.md
+++ b/helm/slurm-operator/README.md
@@ -63,6 +63,7 @@ Kubernetes: `>= 1.29.0-0`
 | operator.slurmclientWorkers | int | `2` | Set the max concurrent workers for the SlurmClient controller. |
 | operator.tokenWorkers | int | `4` | Set the max concurrent workers for the Token controller. |
 | operator.tolerations | list | `[]` | Tolerations for pod assignment. Ref: https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/ |
+| operator.topologySpreadConstraints | list | `[]` | Topology spread constraints for pod assignment. Prefer scheduling replicas across failure domains (nodes, zones, ...) when running in HA. Ref: https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/ |
 | priorityClassName | string | `""` | Set the priority class to use. Ref: https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/#priorityclass |
 | propagatedNodeConditions | list | `[]` | List of Kubernetes Node Conditions, by type, to propagate to the Slurm node drain reason. Ref: https://kubernetes.io/docs/reference/node/node-status/#condition |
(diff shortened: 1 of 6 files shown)


vivian-hafener:
Good morning @giuliocalzo,

I'm going to squash these two commits together - README re-generation gets triggered by running `pre-commit run` prior to committing a change, so we expect generated README changes to be contained within the same commit as the changes that triggered them.

Best,
Vivian Hafener

giuliocalzo:
thanks @vivian-hafener sorry for the multiple PRs

vivian-hafener:
This has been merged here: https://github.com/SlinkyProject/slurm-operator/commit/280763a59a0dd73b0fa0fa668392b429245bd220

PR #183, opened 2026-04-27
fix(builder): honor service.nodePort for restapi, controller, accounting

### LINKED ISSUE #123: [Bug]: slurm key and jwt secret cannot be disabled
## Description
slurm key and jwt secrets cannot be skipped, only when the `.Values.accounting.external` `.Values.controller.external` is set to true. This limitation can lead to issues when using tools like ArgoCD, especially if the Slurm key is generated outside of the Helm release process.

In the documentation, it states: `If empty, one will be generated and used.` However, even when a value is set, the secret continues to be generated.

## Steps to Reproduce

```
helm template helm/slurm \
  --set slurmKeyRef.name=test --set slurmKeyRef.key=testkey \ 
  --set jwtHs256KeyRef.name=test --set jwtHs256KeyRef.key=testkey | grep "kind: Secret" | wc -l
       2
```

## Expected Behavior

```
helm template helm/slurm \
  --set slurmKeyRef.name=test --set slurmKeyRef.key=testkey \ 
  --set jwtHs256KeyRef.name=test --set jwtHs256KeyRef.key=testkey | grep "kind: Secret" | wc -l
       0
```

martbhell:
The chart documents `restapi.service.nodePort`, `controller.service.nodePort`, and `accounting.service.nodePort` (helm/slurm/values.yaml lines 297, 397, 509), but the corresponding ServicePort builders never set NodePort. Setting the field in helm values made it through to the CR's spec.service.nodePort, but the operator silently dropped it when building the Service — kube-apiserver then auto-allocated a random port, and the field was effectively a no-op.

Mirror the pattern already used by loginbuilder/login_service.go: pass spec.NodePort through to the corev1.ServicePort. Add a focused unit test per builder that asserts the propagation.

loginbuilder/login_service.go already has this; loginsets.slinky.service.nodePort in helm values has always worked.

## Checklist

- [x] I have read
  and the

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 All notable changes to this project will be documented in version specific
 files.
 
+- [CHANGELOG-1.2.md](./CHANGELOG/CHANGELOG-1.2.md)
 - [CHANGELOG-1.1.md](./CHANGELOG/CHANGELOG-1.1.md)
 - [CHANGELOG-1.0.md](./CHANGELOG/CHANGELOG-1.0.md)
 - [CHANGELOG-0.4.md](./CHANGELOG/CHANGELOG-0.4.md)
--- a/CHANGELOG/CHANGELOG-1.2.md
+++ b/CHANGELOG/CHANGELOG-1.2.md
@@ -1,0 +1,5 @@
+## v1.2.0
+
+### Fixed
+
+- Honor `service.nodePort` on RestApi, Controller, and Accounting CRs.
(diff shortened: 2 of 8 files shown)


vivian-hafener:
Good morning @martbhell.

Thanks for the contribution, I will review this. Just a note for future reference, we never modify the CHANGELOGs directly, we generally use a `Changelog:` commit trailer. No need to fix that now, I will handle that myself during review.

Best,
Vivian Hafener

martbhell:
Thanks! As a note, this suggestion came from me using Claude code to access slurmrestd api from a node (on the same L2/subnet as a slinky) that hasnt joined the k8s cluster.

vivian-hafener:
Good morning @martbhell,

I've modified your commit to only make changes to the `*_service.go` files, and added a commit of my own (on our internal branch for this PR) that rewrites the tests to follow the established pattern of one test for each BuildComponentService function, instead of adding new test functions just for NodePort.

Best,
Vivian Hafener

vivian-hafener:
This has been merged: https://github.com/SlinkyProject/slurm-operator/commit/7f49b118d32b170cf076fb184a91f0b3ba6fa0da https://github.com/SlinkyProject/slurm-operator/commit/c2d09ff12dab7a077174271492b151b661589f11

PR #187, opened 2026-04-30
Expose nodeSelector for operator and webhook

faganihajizada:
## Summary

The Helm chart did not document or wire `nodeSelector` for either the operator or webhook Deployment. Helm parsed `--set operator.nodeSelector.foo=bar` (and the webhook equivalent) into `.Values` without complaint and silently dropped the value at render time, since no template referenced it and the chart has no `values.schema.json` for validation.

This PR wires `.Values.operator.nodeSelector` and `.Values.webhook.nodeSelector` through `templates/operator/deployment.yaml` and `templates/webhook/deployment.yaml` using the same `{{- with ... }}` / `toYaml | nindent 8` idiom already used for `affinity` and `tolerations`, documents the field in `values.yaml` with wording aligned to the sibling `helm/slurm` chart, regenerates the matching rows in the values table of `helm/slurm-operator/README.md`, and adds helm-unittest cases that assert the rendered field for both deployments.

### Reproduction

    helm template slinky-slurm-op helm/slurm-operator \
      --set operator.nodeSelector.foo=bar \
      --set webhook.nodeSelector.baz=qux | grep nodeSelector
    # (no output — value silently dropped)

## Checklist

## Breaking Changes

Not a breaking API change — `nodeSelector` was never a documented or supported field on either deployment, and default renders are unchanged (the field is omitted unless the user sets it).

There is a latent behavior change worth flagging: any operator who had been setting `operator.nodeSelector` or `webhook.nodeSelector` in their values file under the assumption it was being honored will now see the constraint take effect. They should verify their value does not pin pods to nodes that no longer match.

## Testing Notes

The new helm-unittest cases (`should add nodeSelector`) sit alongside the existing `should add affinity` / `should add tolerations` tests in both `tests/operator_deployment_test.yaml` and `tests/webhook_deployment_test.yaml`, following the same shape and depth.

### DIFF
--- a/helm/slurm-operator/README.md
+++ b/helm/slurm-operator/README.md
@@ -52,6 +52,7 @@ Kubernetes: `>= 1.29.0-0`
 | operator.loginsetWorkers | int | `4` | Set the max concurrent workers for the LoginSet controller. |
 | operator.metricsPort | int | `8080` | Set the port used by the metrics server. Value of "0" will disable it. |
 | operator.namespaces | string | `""` | Comma-separated list of namespaces the operator will watch. If empty, all namespaces are watched. |
+| operator.nodeSelector | object | `{}` | Node label selector for pod assignment. Ref: https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#nodeselector |
 | operator.nodesetWorkers | int | `4` | Set the max concurrent workers for the NodeSet controller. |
 | operator.pdb.enabled | bool | `false` | Enable PodDisruptionBudget. |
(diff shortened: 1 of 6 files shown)


faganihajizada:
merged https://github.com/SlinkyProject/slurm-operator/commit/c4e8959607e79f01d4c2ad665b3db18d6d9d8658

vivian-hafener:
Merged: https://github.com/SlinkyProject/slurm-operator/commit/c4e8959607e79f01d4c2ad665b3db18d6d9d8658

PR #188, opened 2026-05-04
fix(webhook): limit pod binding webhook scope

tsungjung411:
## Summary

Limit the scope of the `podsbinding-v1.kb.io` mutating webhook by adding an `objectSelector` that matches only Slurm worker Pods labeled with `app.kubernetes.io/name: slurmd`.

This prevents the webhook from being invoked for unrelated Pods during Pod binding admission.

Addresses #189.

**Changes:**

1. Add `objectSelector` to the `podsbinding-v1.kb.io` webhook in the Helm template:

   ```yaml
   objectSelector:
     matchLabels:
       app.kubernetes.io/name: slurmd
   ```

2. Add a Helm unit test to verify the `objectSelector` is rendered for the Pod binding webhook.

3. Keep the existing Go-level filtering as a defensive fallback.

**Why this is needed:**

The current webhook is registered for `pods/binding`, so Pod binding requests can reach the webhook before the Go handler filters out non-worker Pods.

Although the Go handler already returns early for non-worker Pods, the request still has to reach the webhook service first. If the webhook service is temporarily unreachable during cluster or node reboot scenarios, unrelated Pod scheduling can be affected.

Adding an admission-level `objectSelector` prevents non-Slurm Pods from invoking the webhook in the first place, reducing the webhook's blast radius while preserving the existing `failurePolicy: Fail` behavior.

## Checklist

* [x] I have read
  and the
* [x] New or existing tests cover these changes (where applicable).
* [x] Documentation is updated if user-visible behavior changes.

## Breaking Changes

None.

This PR does not change the webhook `failurePolicy`; it remains `Fail`.

## Testing Notes

1. **Generate manifests:**

   ```bash
   make manifests
   ```

2. **Generate Helm docs:**

   ```bash
   make helm-docs
   ```

3. **Validate Helm chart:**

   ```bash
   make helm-validate
   ```

4. **Run Helm unit tests:**

   ```bash
   make helm-unittest
   ```

5. **Run Go tests:**

   ```bash
   make test
   ```

   Result:

   ```text
   total: (statements) 75.6%
   go tool cover -html cover.out -o cover.html
   ```

6. **Run pre-commit checks:**

   ```bash
   pre-commit run --all-files
   ```

7. **Verify Helm template rendering:**

   ```bash
   helm template slurm-operator ./helm/slurm-operator | grep -A5 "objectSelector"
   ```

   Expected output includes:

   ```yaml
   objectSelector:
     matchLabels:
       app.kubernetes.io/name: slurmd
   ```

## Additional Context

The related issue describes a cluster or node reboot scenario where the webhook Service can become temporarily unreachable while the cluster networking stack is recovering. With the previous configuration, the `pods/binding` webhook could be invoked for Pods outside of Slurm worker Pods, increasing the blast radius of that temporary webhook unavailability.

The Go handler at `internal/webhook/pod_binding_webhook.go` already filters non-worker Pods with an early `return nil`. This PR moves that filtering boundary earlier, from webhook handler logic to Kubernetes admission selection.

The Go-level check is preserved as a safety net.

Related discussion and full deadlock analysis: #189

### DIFF
--- a/helm/slurm-operator/templates/webhook/webhook.yaml
+++ b/helm/slurm-operator/templates/webhook/webhook.yaml
@@ -263,6 +263,9 @@ webhooks:
           values:
             - kube-system
             - {{ include "slurm-operator.namespace" . }}
+    objectSelector:
+      matchLabels:
+        app.kubernetes.io/name: slurmd
     admissionReviewVersions:
       - v1
     clientConfig:
--- a/helm/slurm-operator/tests/webhook_webhook_test.yaml
+++ b/helm/slurm-operator/tests/webhook_webhook_test.yaml
@@ -87,6 +87,9 @@ tests:
       - equal:
           path: webhooks[0].rules[0].resources[0]
           value: pods/binding
+      - equal:
+          path: webhooks[0].objectSelector.matchLabels["app.kubernetes.io/name"]
+          value: slurmd
 
   - it: should include cert-manager annotations when certManager is enabled
     documentSelector:
--- a/internal/webhook/pod_binding_webhook.go
+++ b/internal/webhook/pod_binding_webhook.go
@@ -37,6 +37,8 @@ func (r *PodBindingWebhook) SetupWebhookWithManager(mgr ctrl.Manager) error {
 // +kubebuilder:rbac:groups="",resources=pods,verbs=get;list;update;patch;watch
 // +kubebuilder:rbac:groups="",resources=pods/binding,verbs=get;list;watch
 // +kubebuilder:webhook:path=/mutate--v1-binding,mutating=true,failurePolicy=fail,matchPolicy=Equivalent,sideEffects=None,groups="",resources=pods/binding,verbs=create,versions=v1,name=podsbinding-v1.kb.io,admissionReviewVersions=v1
+// NOTE: objectSelector (matchLabels: app.kubernetes.io/name=slurmd) is manually
+// maintained in config/webhook/manifests.yaml and helm template.
 
 var _ admission.Defaulter[*corev1.Binding] = &PodBindingWebhook{}
 


config/webhook/manifests.yaml
```diff
@@ -11,9 +11,12 @@ webhooks:
       name: webhook-service
       namespace: system
       path: /mutate--v1-binding
-  failurePolicy: Fail
+  failurePolicy: Ignore
```
vivian-hafener: I do not agree with the decision to change the default `failurePolicy` from `Fail` to `Ignore`. `Fail` is a sensible default given the context within which we expect the webhook to run.

helm/slurm-operator/templates/webhook/webhook.yaml
```diff
@@ -273,7 +276,11 @@ webhooks:
         namespace: {{ include "slurm-operator.namespace" . }}
         name: {{ include "slurm-operator.webhook.name" . }}
         path: /mutate--v1-binding
-    failurePolicy: Fail
+    {{- $podBindingFP := .Values.webhook.podBinding.failurePolicy | default "Ignore" }}
+    {{- if not (or (eq $podBindingFP "Fail") (eq $podBindingFP "Ignore")) }}
```
vivian-hafener: https://github.com/SlinkyProject/slurm-operator/pull/188/changes#r3184059236

helm/slurm-operator/values.yaml
```diff
@@ -173,6 +173,13 @@ webhook:
   # -- Comma-separated list of namespaces the webhook will watch.
   # If empty, all namespaces are watched.
   namespaces: ""
+  # pods/binding webhook configurations.
+  podBinding:
+    # -- Set the failure policy for the pods/binding MutatingWebhook.
+    # "Ignore" (default) prevents cluster-wide scheduling outages if the webhook is unreachable.
+    # "Fail" ensures strict scheduling consistency for Slurm worker Pods.
+    # Ref: https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#failure-policy
```
vivian-hafener: https://github.com/SlinkyProject/slurm-operator/pull/188/changes#r3184059236

vivian-hafener (changes requested):
This needs work. Neither commit passes our `commitlint` pre-commit hook or our `make-helm-docs` hook (changes were requested by the hook but not made). The first commit fails the `Make Manifests` hook (changes were requested by the hook but not made). The second commit fails the `helm unittest` hook. This latter failure indicates a logical failure in your implementation

Please comply with our contributing guidelines: https://github.com/SlinkyProject/slurm-operator/blob/main/CONTRIBUTING.md#code-contributions

Additionally, please test and review your changes prior to submitting them.

tsungjung411:
Thanks for the review. I have updated the PR to address the feedback:

- Removed the `failurePolicy` change entirely; it remains `Fail`.
- Squashed/amended the PR into a single conventional commit with Signed-off-by.
- Re-ran generated files and removed the manual change to `config/webhook/manifests.yaml`.
- Added a Helm unit test for the new `objectSelector`.
- Verified the PR now only changes:
  - `helm/slurm-operator/templates/webhook/webhook.yaml`
  - `helm/slurm-operator/tests/webhook_webhook_test.yaml`
  - `internal/webhook/pod_binding_webhook.go`

I also ran:

```bash
make manifests
make helm-docs
make helm-validate
make helm-unittest
make test
pre-commit run --all-files
```

The PR is now scoped only to limiting the pod binding webhook admission scope using `objectSelector`.

tsungjung411:
Hi @vivian-hafener,

After further testing, I discovered that `objectSelector` does **not** work as expected on `pods/binding` subresource webhooks. Here's what I found:

### The Problem with `objectSelector` on `pods/binding`

When the Kubernetes scheduler binds a Pod to a Node, it creates a **Binding object**:

```go
b := &v1.Binding{
    ObjectMeta: metav1.ObjectMeta{Namespace: p.Namespace, Name: p.Name, UID: p.UID},
    Target:     v1.ObjectReference{Name: nodeName},
}
```

The Binding object only carries `Name`, `Namespace`, and `UID` — it does **not** inherit the Pod's labels. Since `objectSelector` evaluates the **admitted object's own labels** (i.e., the Binding, not the parent Pod), the selector `matchLabels: app.kubernetes.io/name: slurmd` will never match any Binding object.

### Test Results

I verified this on a live cluster by running the following commands:
> ```
> $ kubectl run test-normal --image=busybox --command -- sleep 3600
> $ kubectl run test-slurmd --image=busybox --labels="app.kubernetes.io/name=slurmd" --command -- sleep 3600
> ```

- **Without `objectSelector`**: webhook intercepts all pod bindings (both `test-normal` and `test-slurmd` pods show up in webhook logs)
- **With `objectSelector`**: webhook intercepts **nothing** — not even actual Slurm worker pods. The webhook becomes completely inactive.

### What I'm Exploring

Given that `objectSelector` is not viable for `pods/binding`, I'm investigating alternative approaches to reduce the webhook's blast radius while keeping `failurePolicy: Fail`:

1. **`matchConditions` (CEL expressions, K8s 1.27+)** — Can filter on `request.name` or `request.namespace`, though it cannot access the parent Pod's labels either. Could potentially filter by pod name prefix pattern if naming conventions are stable.

2. **Broader `namespaceSelector`** — Expanding the excluded namespaces to cover more system-critical workloads.

3. **Existing code-level filtering** — The Go handler already returns early for non-worker Pods (line 55-56 in `pod_binding_webhook.go`). The remaining concern is webhook reachability during cluster bootstrap/reboot scenarios (as described in #189).

I'll update this PR once I have a working alternative, or close it if the current `namespaceSelector` + code-level filtering is deemed sufficient. Would appreciate your thoughts on which direction makes the most sense for the project.
```

PR #194, opened 2026-05-12
fix(login): raise RSA SSH host key default to 4096 bits

giuliocalzo:
## Summary

- `crypto.NewKeyPair` defaulted RSA keys to 1024 bits, and `BuildLoginSshHostKeys` was relying on that default to generate the login pod's RSA SSH host key. 1024-bit RSA is below modern security baselines (NIST SP 800-57 recommends 3072 bits for keys remaining secure past 2030).
- Introduces `crypto.DefaultRsaBitLength = 4096` and uses it as the new default in `NewKeyPair`.
- Pins the length explicitly at the login-secret call site (`crypto.WithRsaLength(crypto.DefaultRsaBitLength)`) so the secure length is documented at both layers and won't silently regress if the default ever changes.

## Test plan

- [x] `go test ./internal/utils/crypto/... ./internal/builder/loginbuilder/...` passes locally (existing `TestNewKeyPair` cases cover RSA default, `WithRsaLength(4096)`, and the rejected `WithRsaLength(256)` insecure-length case).
- [ ] CI green.
- [x] Manual verification: deploy a `LoginSet`, exec into the resulting Secret, and confirm `ssh-keygen -lf ssh_host_rsa_key` reports 4096 bits.

### DIFF
--- a/internal/builder/loginbuilder/login_secret.go
+++ b/internal/builder/loginbuilder/login_secret.go
@@ -16,7 +16,10 @@ import (
 )
 
 func (b *LoginBuilder) BuildLoginSshHostKeys(loginset *slinkyv1beta1.LoginSet) (*corev1.Secret, error) {
-	keyPairRsa, err := crypto.NewKeyPair(crypto.WithType(crypto.KeyPairRsa))
+	keyPairRsa, err := crypto.NewKeyPair(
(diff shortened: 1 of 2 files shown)


SkylerMalinowski:
Merged: 8ecab27c3970f62961f6dd8b3c074a99c701c062

PR #195, opened 2026-05-18
fix: show controller/restapi connection info

faganihajizada:
## Summary

Fix a templating bug in `helm/slurm/templates/NOTES.txt` where the post-install connection instructions for `slurmctld` and `slurmrestd` were silently hidden on a default install.

The outer `{{- if ne ... }}` guards used `"None"` as the `dig` default for `service.spec.type`. With the values.yaml default of `spec: {}`, `dig` returned the sentinel and skipped both blocks even though Kubernetes still creates a ClusterIP service that users need to know how to reach.
Switch the outer guards to `dig ... ""` to match the pattern already used by the inner `LoadBalancer` / `ExternalName` / `clusterIP` branches in the same file. Default installs now render the connection instructions; explicitly setting `service.spec.type: None` still opts out.

## Checklist

- [ ] New or existing tests cover these changes (where applicable).

## Breaking Changes

N/A

## Testing Notes

Rendered NOTES with `helm install --dry-run` against the chart:
- **Defaults** (`spec: {}`): controller and restapi sections now appear with `jsonpath='{.spec.clusterIP}'`, as expected.
- **Explicit opt-out** (`--set controller.service.spec.type=None --set restapi.service.spec.type=None`): both sections suppressed.
- **LoadBalancer / ExternalName**: inner branches unchanged; renders the matching jsonpath.

### DIFF
--- a/helm/slurm/templates/NOTES.txt
+++ b/helm/slurm/templates/NOTES.txt
@@ -63,7 +63,7 @@ ssh via the Slurm login ({{ $name }}) service:
 {{- end }}{{- /* $loginset.enabled */}}
 {{- end }}{{- /* range $key, $loginset := $.Values.loginsets */}}
 
-{{- if ne (.Values.controller | dig "service" "spec" "type" "None" | lower) ("None" | lower) }}
+{{- if ne (.Values.controller | dig "service" "spec" "type" "" | lower) ("None" | lower) }}
 
 The Slurm controller service:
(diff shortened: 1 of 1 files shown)


vivian-hafener:
This was merged here: https://github.com/SlinkyProject/slurm-operator/commit/941fd7674a4c8d33052e115830542da417e584e9

PR #192, opened 2026-05-08
fix(webhook): replace objectSelector with namespaceSelector filtering to prevent webhook deadlock (#189)

tsungjung411:
## Summary

Follow-up to: [fix(webhook): limit pod binding webhook scope #188](https://github.com/SlinkyProject/slurm-operator/pull/188)

Replace the ineffective `objectSelector` with a configurable `namespaceSelector` inclusion list for the `podsbinding-v1.kb.io` webhook. This reduces the webhook's blast radius by allowing users to limit interception to only namespaces where Slurm clusters are deployed, preventing cluster-wide scheduling deadlocks during node reboots.

Addresses #189.

## Problem

The previous approach using `objectSelector` with `matchLabels: app.kubernetes.io/name: slurmd` does **not work** for `pods/binding` subresource webhooks. The Kubernetes Binding object does not inherit the parent Pod's labels, so the selector never matches — causing the webhook to become completely inactive (not just filtered).

With the webhook intercepting all `pods/binding` requests in non-excluded namespaces and `failurePolicy: Fail`, any webhook unavailability (cluster reboot, CNI recovery, rolling update) blocks all Pod scheduling cluster-wide.

## Solution

Add a configurable `webhook.podBinding.includedNamespaces` field in `values.yaml`:

```yaml
webhook:
  podBinding:
    # When set, only pod bindings in these namespaces will be intercepted.
    # Example: ["slurm", "slurm-prod"]
    includedNamespaces: []
```

When configured, this adds a `kubernetes.io/metadata.name: In [...]` expression to the existing `namespaceSelector`, restricting the webhook to only the specified namespaces. The existing `NotIn` exclusion for `kube-system` and operator namespace is preserved.

## Changes

- **`helm/slurm-operator/values.yaml`**: Add `webhook.podBinding.includedNamespaces` configuration (default: `[]`)
- **`helm/slurm-operator/templates/webhook/webhook.yaml`**: Remove ineffective `objectSelector`; add conditional `namespaceSelector` In expression
- **`helm/slurm-operator/tests/webhook_webhook_test.yaml`**: Remove objectSelector test; add test for includedNamespaces rendering

## Why `namespaceSelector` instead of `objectSelector`

| Approach | Works for `pods/binding`? | Reason |
|----------|--------------------------|--------|
| `objectSelector` (labels) | ❌ | Binding object has no Pod labels |
| `matchConditions` (CEL) | ⚠️ Partial | Cannot access Pod labels; pod name prefix depends on Helm release name |
[...]

PR #196, opened 2026-05-19
feat(helm): parametrize webhook failurePolicy/matchPolicy and default to Ignore

giuliocalzo:
## Summary

- Expose `webhook.failurePolicy` and `webhook.matchPolicy` as Helm values, applied uniformly to every entry in the generated `ValidatingWebhookConfiguration` and the pods/binding `MutatingWebhookConfiguration`.
- Default `webhook.failurePolicy` to `Ignore` (was hard-coded `Fail`); `webhook.matchPolicy` stays at `Equivalent`.
- Documented in `values.yaml` and `README.md`.
- Refactored `tests/webhook_webhook_test.yaml` so the default-rendered manifest is locked in via `matchSnapshot` and each remaining unit test only asserts the single parameter it changes.

## Root cause

The chart previously hard-coded:

```yaml
failurePolicy: Fail
matchPolicy: Equivalent
```

on every webhook entry, including the cluster-wide `podsbinding-v1.kb.io` mutating webhook that runs on every pod's `Bind` step. With `failurePolicy=Fail`, any window in which the webhook Service has no ready endpoints — rollouts, evictions, node drains, image pulls, OOMKill, PDB-induced downtime — causes the kube-scheduler's `DefaultBinder` plugin to fail and **all** new pods in **all** namespaces get stuck in `Pending` with events like:

```text
Warning FailedScheduling default-scheduler
  running Bind plugin "DefaultBinder": Internal error occurred:
  failed calling webhook "podsbinding-v1.kb.io":
  Post "https://slurm-operator-webhook.slurm-operator.svc:443/mutate--v1-binding?timeout=10s":
  no endpoints available for service "slurm-operator-webhook"
```

This effectively makes the slurm-operator webhook a cluster-wide control-plane dependency for unrelated workloads (DNS, ingress, monitoring, the operator's own restart, etc.), and the failure mode is self-reinforcing: if the webhook pod itself can't be rescheduled, nothing else can either.

## Fix

- Default `failurePolicy: Ignore` so an unavailable webhook is fail-open and does **not** block pod scheduling. The pods/binding mutation is best-effort scheduling guidance for Slurm-managed workloads; skipping it during webhook downtime is preferable to halting the entire cluster.
- Strict enforcement remains available with `--set webhook.failurePolicy=Fail` for operators who run the webhook in HA and accept the availability/consistency trade-off.
- `matchPolicy` is exposed as a knob (default unchanged at `Equivalent`) so users can flip it to `Exact` if their admission-policy posture requires it.

## Files changed

- `helm/slurm-operator/values.yaml` — new `webhook.failurePolicy` / `webhook.matchPolicy` values with rationale comments.
- `helm/slurm-operator/templates/webhook/webhook.yaml` — replaced the 7 hard-coded pairs with `.Values.webhook.failurePolicy` / `.Values.webhook.matchPolicy`.
- `helm/slurm-operator/README.md` — value table updated.
- `helm/slurm-operator/tests/webhook_webhook_test.yaml` — restructured: a single `matchSnapshot` test pins the default-rendered ValidatingWebhookConfiguration + MutatingWebhookConfiguration (deterministic because the chart-default `certManager.enabled=true` omits the inline TLS cert), and the remaining tests use `equal`/`exists` to assert only the parameter each one overrides — `failurePolicy`/`matchPolicy`, `timeoutSeconds`, and the `certManager.enabled=false` path (caBundle injection + TLS Secret render).
- `helm/slurm-operator/tests/__snapshot__/webhook_webhook_test.yaml.snap` — new committed snapshot.

## Test plan

- [x] `helm unittest helm/slurm-operator` — 69 tests / 12 snapshots pass locally.
- [x] Re-running the snapshot test multiple times produces identical output (no flake from `genCA` / `genSignedCert`, because the default path uses cert-manager and never generates inline material).
- [x] `helm template t helm/slurm-operator --set webhook.failurePolicy=Ignore --set webhook.matchPolicy=Exact --set certManager.enabled=false` renders the overrides on all 7 webhook entries.
- [ ] CI green.
- [x] Manual: in a cluster with the chart installed, scale the webhook deployment to 0 and confirm new pods still schedule (with the warning event suppressed) instead of getting stuck in `Pending`.

## Backwards compatibility

This changes the default failure mode for the chart from fail-closed to fail-open. Users that rely on strict admission of slurm-operator CRs and the pods/binding mutation should set `webhook.failurePolicy: Fail` in their values overrides on upgrade.

### DIFF
--- a/helm/slurm-operator/README.md
+++ b/helm/slurm-operator/README.md
@@ -68,11 +68,13 @@ Kubernetes: `>= 1.29.0-0`
 | propagatedNodeConditions | list | `[]` | List of Kubernetes Node Conditions, by type, to propagate to the Slurm node drain reason. Ref: https://kubernetes.io/docs/reference/node/node-status/#condition |
 | webhook.affinity | object | `{}` | Affinity for pod assignment. Ref: https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#affinity-and-anti-affinity |
 | webhook.enabled | bool | `true` | Enable the webhook. |
+| webhook.failurePolicy | string | `"Ignore"` | Action taken when the admission webhook is unreachable or returns an error. Applied to every webhook entry in the generated ValidatingWebhookConfiguration and MutatingWebhookConfiguration. Default is "Ignore" so that an unavailable webhook (e.g. during rollout, eviction, or node drain) does not block unrelated cluster operations such as pod scheduling via the pods/binding mutating webhook. Set to "Fail" to require strict enforcement at the cost of cluster availability when the webhook is down. Ref: https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#failure-policy |
 | webhook.healthPort | int | `8081` | Set the port used for health checks. |
 | webhook.image | object | `{"repository":"ghcr.io/slinkyproject/slurm-operator-webhook","tag":""}` | The image to use, `${repository}:${tag}`. Ref: https://kubernetes.io/docs/concepts/containers/images/#image-names |
 | webhook.imagePullPolicy | string | `"IfNotPresent"` | Set the image pull policy. |
 | webhook.leaderElection | bool | `true` | Enable leader election for slurm-operator-webhook |
 | webhook.logLevel | string | `"info"` | Set the log level by string (e.g. error, info, debug) or number (e.g. 1..5). |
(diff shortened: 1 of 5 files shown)


vivian-hafener:
Hi @giuliocalzo,

Thanks for the contribution. I will review this.

Related: https://github.com/SlinkyProject/slurm-operator/pull/192

As this PR does expose the `failurePolicy` for user modification, I think that changing the default to `Ignore` is reasonable.

Best,
Vivian Hafener

vivian-hafener:
Good morning,

Thank you for the contribution.

This has been merged:
* https://github.com/SlinkyProject/slurm-operator/commit/03fa998f436a4194ced8ee2cf27f4d4657f04f70
* https://github.com/SlinkyProject/slurm-operator/commit/b24d657068e3840954e87611dad2d9e8aaa58425
* https://github.com/SlinkyProject/slurm-operator/commit/1655a8c69ee1db54d7e994075e90d11c0e4faee3

Best,
Vivian Hafener

PR #198, opened 2026-05-21
feat(docs): Add GKE TPU Slinky tutorial for JAX

michaelasp:
## Checklist

- [X] I have read
  [CONTRIBUTING.md](https://github.com/SlinkyProject/slurm-operator/blob/main/CONTRIBUTING.md)
  and the
  [Code of Conduct](https://github.com/SlinkyProject/slurm-operator/blob/main/CODE_OF_CONDUCT.md).
- [X] New or existing tests cover these changes (where applicable).
- [X] Documentation is updated if user-visible behavior changes.

## Breaking Changes

N/A

## Testing Notes
N/A

## Additional Context
Adds documentation on how to use slurm/slinky with TPUs on GKE. This is my first time contributing to slinky so let me know if there's anything else that needs to be done!

## Verification
This was all verified on a GKE cluster running TPU nodes.

PR #199, opened 2026-05-21
fix(client): Add check if CRD is installed to prevent infinite error loops on metrics

### LINKED ISSUE #123: [Bug]: slurm key and jwt secret cannot be disabled

michaelasp:
## Summary
Ran into this while testing #198, if the metrics CRDs are not installed then the operator will get stuck in an infinite loop here. Updated the delete function to ensure that it returns nil if the CRD is not installed on the cluster.

## Checklist

- [x] I have read
  and the

## Breaking Changes
None

## Testing Notes
Added test case for this scenario, in an actual cluster create a slurm cluster without the ServiceMonitor CRD installed.

## Additional Context
N/A

### DIFF
--- a/internal/utils/objectutils/delete.go
+++ b/internal/utils/objectutils/delete.go
@@ -13,6 +13,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	policyv1 "k8s.io/api/policy/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/client-go/tools/events"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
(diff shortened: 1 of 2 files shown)


vivian-hafener:
Good afternoon @michaelasp,

I believe that I resolved this bug here: https://github.com/SlinkyProject/slurm-operator/commit/d5812e018477abf2491c9503e340d4638614e7f5

Please let me know if the fix provided in that commit does not work in your environment.

Best,
Vivian Hafener

michaelasp:
> Good afternoon @michaelasp,
> 
> I believe that I resolved this bug here: [d5812e0](https://github.com/SlinkyProject/slurm-operator/commit/d5812e018477abf2491c9503e340d4638614e7f5)
> 
> Please let me know if the fix provided in that commit does not work in your environment.
> 
> Best, Vivian Hafener

Ah yeah I was running this on 1.1.0 which probably doesn't have that fix. I can close this since I didn't test against main.

PR #201, opened 2026-05-28
feat(restapi): add helm PodDisruptionBudget

giuliocalzo:
## Summary

- Add an optional `PodDisruptionBudget` for the slurmrestd (`RestApi`) pods in the `slurm` Helm chart.
- The PDB selector matches the labels the operator applies to the slurmrestd pods (`app.kubernetes.io/name=slurmrestd` and `app.kubernetes.io/instance=<release fullname>`), via a new `slurm.restapi.selectorLabels` helper.
- Gated behind `restapi.pdb.enabled` and only rendered when `restapi.replicas > 1` (a PDB on a single-replica Deployment would block all voluntary evictions).
- Supports `minAvailable` (default `1`) and `maxUnavailable` (takes precedence when set).

This is implemented purely at the Helm level; no CRD/operator (Go) changes.

## Test plan

- [x] `helm template` renders the PDB with correct name, namespace, labels and selector.
- [x] `helm unittest` suite `tests/restapi_pdb_test.yaml` (6 cases): disabled, single-replica skip, metadata/selector/spec, custom minAvailable, maxUnavailable precedence, namespaceOverride.
- [x] `helm lint` / `make helm-validate` pass.
- [x] All pre-commit hooks green (incl. helm-docs, helm-validate, helm-unittest).


SkylerMalinowski:
Merged: 7c85b3d8

PR #203, opened 2026-06-03
feat: sync Slurm node features from k8s node annotation

faganihajizada:
## Summary

Add support for syncing Slurm node features from a Kubernetes Node annotation.

When a NodeSet pod is running on a Kubernetes Node annotated with `features.slinky.slurm.net/spec` (a comma-separated list of Slurm features), the operator merges that list with the per-NodeSet baseline and applies it to the registered Slurm node's available and active features. Jobs can then target those nodes with `sbatch --constraint=<feature>`.

The driving use case is tagging each node with a dynamically discovered E/W leaf-switch ID (e.g. `nn-75bfcf47ca3e4f7dc`) so users can constrain jobs to a single switch for network locality, but the mechanism is general: the annotation value is an opaque feature set and is not interpreted by the operator.

How it works:

- New well-known node annotation `features.slinky.slurm.net/spec` and `FeaturesPrefix` (`api/v1beta1/well_known.go`).
- A new `syncSlurmFeatures` reconcile step reads the annotation off the Kubernetes Node each NodeSet pod is bound to. For Nodes that carry the annotation it merges it with `BaselineFeatures` (the NodeSet name or hostname override plus any `Feature=`/`Features=` entries from `Spec.ExtraConf`) and pushes the sorted, de-duplicated union to the Slurm node via the REST API, setting both `Features` (available) and `FeaturesAct` (active). Nodes without the annotation are left untouched, since slurmd already registered them with the baseline.
- The existing `ExtraConf` parsing is extracted into shared, exported helpers (`ParseExtraConf`, `BaselineFeatures`, `SortedDedup`) in `workerbuilder`, so the slurmd build path and the new sync path use one contract.
- The NodeSet validation webhook validates `Spec.ExtraConf` at admission using the same parser.

Delivery is through the reconcile loop only: `NodeEventHandler` already re-enqueues a NodeSet when a Node's metadata changes. The component that discovers topology and writes the annotation (e.g. a daemon querying the CSP topology API) is out of scope for this change; the operator only propagates the annotation it finds.

## Checklist

## Breaking Changes

The NodeSet validation webhook (when `webhook.enabled` is set, the slurm-operator chart default) now rejects `Spec.ExtraConf` whose contents are not whitespace-separated `Key=Value` pairs, on both create and update.

This is not a behavioral regression. Previously, malformed `ExtraConf` wedged the affected NodeSet's reconcile: building its slurmd pod parsed the same string and `panic`ed. controller-runtime recovers the panic (the manager process does not crash), but that NodeSet never gets its worker pods built and the reconcile keeps retrying. This change does two things. It surfaces the problem as a clear admission error at apply time (via the webhook, using the exact same parser), and it makes the build path degrade to the baseline `Features=<name>` instead of panicking, so a NodeSet that bypasses the webhook (a separate, optional deployment) no longer wedges its reconcile. A NodeSet created before this validation that carries malformed `ExtraConf` is rejected on its next update until the `ExtraConf` is corrected.

## Testing Notes

Automated (`make test`, `go vet`, `make golangci-lint` all pass):

- `workerbuilder`: `TestParseExtraConf` (valid, CSV, case-normalisation, `Feature`/`Features` distinct keys, whitespace tolerance, malformed errors), `TestBaselineFeatures` (name-only, `Feature=`, `Features=` CSV, dedup, hostname override, malformed-falls-back-to-baseline, empty-hostname-falls-back-to-name), `TestSlurmdConfArgs` (full `--conf` output, including malformed-degrades-to-baseline).
- `slurmcontrol`: `Test_realSlurmControl_UpdateNodeFeatures` asserts both `Features` and `ActiveFeatures` land, stale features are replaced, an already-in-sync node issues zero updates (order-insensitive skip), an unregistered node is tolerated, and a non-tolerated update error is surfaced.
- `nodeset`: `TestNodeSetReconciler_syncSlurmFeatures` (pending pod, allocated with annotation, allocated-without-annotation skips the update, `=`-bearing feature value) with a counting interceptor asserting no update is issued where expected.
- `webhook`: `nodeset_webhook_test.go` denies malformed and admits valid `ExtraConf`.

Manual (kind cluster, slurm-operator + slurm chart):

```sh
# Annotate the Kubernetes node a worker pod runs on:
kubectl annotate node <node> features.slinky.slurm.net/spec=nn-75bfcf47ca3e4f7dc,GPU

# After one reconcile, both feature lists hold the sorted union of the baseline
# (<nodeset-name>) and the annotation:
kubectl exec -n slurm <controller-pod> -c slurmctld -- \
  scontrol show node <slurm-node> | grep -E 'AvailableFeatures|ActiveFeatures'
# AvailableFeatures=GPU,nn-75bfcf47ca3e4f7dc,<nodeset-name>
# ActiveFeatures=GPU,nn-75bfcf47ca3e4f7dc,<nodeset-name>

# A constraint job targets the tagged node:
sbatch --constraint=nn-75bfcf47ca3e4f7dc ...

# Admission rejects malformed extraConf:
kubectl patch nodeset <name> --type=merge -p '{"spec":{"extraConf":"BadFormat"}}'
# rejected: invalid extraConf: malformed --conf item "BadFormat": expected Key=Value
```

Verified end-to-end on the current code (kind, slurm-operator + slurm chart, webhook enabled, NodeSet `extraConf: Feature=rdma`, Slurm node `slinky-0`):

- Propagation: annotating the node `nn-cccc3333,GPU` set both `AvailableFeatures` and `ActiveFeatures` to `GPU,nn-cccc3333,rdma,slinky` (the sorted union of the baseline and the annotation) in a single REST update.
- Re-annotation: changing the annotation to `nn-dddd4444,GPU` replaced the old switch ID while preserving the `rdma,slinky` baseline.
- Skip when unannotated: removing the annotation issued no update (the operator logged none) and the previously applied features lingered rather than reverting to the baseline, matching the documented opt-in behavior; re-adding the annotation re-applied them.
- Admission: a malformed `extraConf` (`BadFormat`) was rejected with `invalid extraConf: malformed --conf item "BadFormat": expected Key=Value`; a valid `extraConf` was admitted.

The build-path degrade on malformed `extraConf` is covered by unit tests (`TestSlurmdConfArgs`); it is not exercised live because, with the webhook enabled, such input is rejected at admission before reaching the build path.

## Additional Context

Out of scope by design: 

- a `POD_FEATURES` env var / startup injection: The reconcile loop is the simpler, image-change-free, and reliable path, and it's required regardless because discovery is async. Startup injection would only shrink a brief eventual-consistency window for nodes already annotated at pod start, which doesn't apply to the async use case and doesn't remove the need for the loop, so my opinin is it's not worth the added cost but open to discuss.

### DIFF
--- a/docs/usage/node-features.md
+++ b/docs/usage/node-features.md
@@ -1,0 +1,105 @@
+# Node Features
+
+## Table of Contents
+
+<!-- mdformat-toc start --slug=github --no-anchors --maxlevel=6 --minlevel=1 -->
+
+- [Node Features](#node-features)
+  - [Table of Contents](#table-of-contents)
+  - [Overview](#overview)
+  - [Kubernetes](#kubernetes)
+  - [Slurm](#slurm)
+  - [Example](#example)
+
+<!-- mdformat-toc end -->
+
+## Overview
+
+The operator can propagate node features from Kubernetes nodes to Slurm nodes
+(those running as NodeSet pods). When a NodeSet pod runs on a Kubernetes node, each
+value in the node's `features.slinky.slurm.net/spec` annotation is applied to the
+registered Slurm node's available and active features under a reserved prefix,
+`k8s/` (for example the annotation `nn-75bfcf47ca3e4f7dc` becomes the Slurm feature
+`k8s/nn-75bfcf47ca3e4f7dc`).
+
+Use the annotation for per-node attributes discovered at runtime, such as a network
+switch ID. NodeSet-wide features (the NodeSet name and any `Feature=`/`Features=`
+entries in `extraConf`) are seeded directly by slurmd at registration via `--conf`
+and are not prefixed.
+
+The operator owns only the `k8s/` namespace. On each reconcile it replaces the
+prefixed features with the current annotation values and preserves every other
+feature, including the NodeSet baseline, `extraConf` features, and features managed
+outside the operator such as those from a Slurm `NodeFeaturesPlugins` plugin or a
+manual `scontrol update`. Removing the annotation clears the node's `k8s/` features
+while leaving the rest intact.
+
+## Kubernetes
+
+Annotate each Kubernetes node with `features.slinky.slurm.net/spec` and a
+comma-separated list of feature names. Each token is used verbatim and applied to
+the Slurm node under the `k8s/` prefix. Only Slurm nodes backed by a NodeSet pod
+scheduled on the node are updated.
+
+For example, the following Kubernetes Node snippet has the Slinky node features
+annotation applied.
+
+```yaml
+apiVersion: v1
+kind: Node
+metadata:
+  name: node0
+  annotations:
+    features.slinky.slurm.net/spec: nn-75bfcf47ca3e4f7dc
+```
+
+The annotation is normally written by an external component that discovers each
+node's features (for example, a daemon querying a cloud provider's network
+topology API). The operator only propagates the annotation it finds.
+
+## Slurm
+
+Slurm node features can be requested by a job with `--constraint`. See the Slurm
+[Features][features-conf] documentation. Because operator-managed features carry
+the `k8s/` prefix, jobs target them as `--constraint=k8s/<feature>`. The operator
+applies each prefixed feature to both available and active features, and because it
+only touches the `k8s/` namespace it does not interfere with features managed by a
+`NodeFeaturesPlugins` plugin.
+
+## Example
+
+Suppose the NodeSet is named `slinky` and sets `extraConf: Features=GPU`. slurmd
+registers the node with the baseline features `slinky` and `GPU` (unprefixed). The
+Kubernetes node where the `slinky-0` NodeSet pod runs is annotated with its
+leaf-switch ID.
+
+```yaml
+---
+apiVersion: v1
+kind: Node
+metadata:
+  name: node0
+  annotations:
+    features.slinky.slurm.net/spec: nn-75bfcf47ca3e4f7dc
+```
+
+When the `slinky-0` NodeSet pod is scheduled onto Kubernetes node `node0`, the
+operator applies the annotation under the `k8s/` prefix, preserving the baseline.
+Slurm then reports both the baseline and the prefixed feature.
+
+```console
+$ scontrol show node slinky-0 | grep -Eo "NodeName=[^ ]+|[ ]*AvailableFeatures=[^ ]+|[ ]*ActiveFeatures=[^ ]+"
+NodeName=slinky-0
+   AvailableFeatures=GPU,k8s/nn-75bfcf47ca3e4f7dc,slinky
+   ActiveFeatures=GPU,k8s/nn-75bfcf47ca3e4f7dc,slinky
+```
+
+A job can then be constrained to that switch.
+
+```sh
+sbatch --constraint=k8s/nn-75bfcf47ca3e4f7dc ...
+```
+
+<!-- Links -->
+
+[features-conf]: https://slurm.schedmd.com/slurm.conf.html#OPT_Features
--- a/internal/builder/workerbuilder/worker_app.go
+++ b/internal/builder/workerbuilder/worker_app.go
@@ -283,48 +283,66 @@ func slurmdArgs(nodeset *slinkyv1beta1.NodeSet, controller *slinkyv1beta1.Contro
 	return args
 }
 
-func slurmdConfArgs(nodeset *slinkyv1beta1.NodeSet) []string {
-	extraConf := []string{}
-	if nodeset.Spec.ExtraConf != "" {
-		extraConf = strings.Split(nodeset.Spec.ExtraConf, " ")
-	}
-
-	name := common.GetSlurmNodeSetName(nodeset)
-	confMap := map[string]string{
-		"Features": name,
-	}
-	for _, item := range extraConf {
+// ParseExtraConf parses a NodeSet's Spec.ExtraConf into a map of normalized keys
+// to their values. The expected shape is space-separated Key=Value pairs; each
+// value can be a comma-separated list (e.g. Features=a,b,c). Keys are title-cased
+// so "feature=a" and "Feature=a" both produce key "Feature".
+//
+// Tokens are split on whitespace, so leading, trailing, and repeated whitespace
+// are tolerated; any token without '=' returns an error. Exported so the NodeSet
+// validation webhook can reject malformed input with the same parser the worker
+// builder uses, keeping the accepted set identical at admission and at build time.
+func ParseExtraConf(extraConf string) (map[string][]string, error) {
+	out := make(map[string][]string, 10)
+	for item := range strings.FieldsSeq(extraConf) {
 		pair := strings.SplitN(item, "=", 2)
-		key := cases.Title(language.English).String(pair[0])
 		if len(pair) != 2 {
-			panic(fmt.Sprintf("malformed --conf item: %v", item))
+			return nil, fmt.Errorf("malformed --conf item %q: expected Key=Value", item)
 		}
-		val := pair[1]
-		if key == "Features" || key == "Feature" {
-			// Slurm treats trailing 's' as optional. We have to
-			// specially handle 'Feature(s)' because we require at
-			// least one feature but the user can request additional.
-			key = "Features"
-		}
-		if ret, ok := confMap[key]; !ok {
-			confMap[key] = val
-		} else {
-			confMap[key] = ret + fmt.Sprintf(",%s", val)
+		key := cases.Title(language.English).String(pair[0])
+		out[key] = append(out[key], strings.Split(pair[1], ",")...)
+	}
+	return out, nil
+}
+
+// baselineFeatures returns the sorted, de-duplicated list of Slurm features baked
+// into slurmd at registration via --conf: the NodeSet name (or hostname override)
+// plus any Feature=/Features= entries from Spec.ExtraConf.
+//
+// A ParseExtraConf error falls back to the bare baseline (just the name) rather
+// than failing, so malformed ExtraConf does not wedge slurmdConfArgs; the NodeSet
+// webhook rejects such input at admission when it is enabled.
+func baselineFeatures(nodeset *slinkyv1beta1.NodeSet) []string {
+	raw := []string{common.GetSlurmNodeSetName(nodeset)}
+	if conf, err := ParseExtraConf(nodeset.Spec.ExtraConf); err == nil {
+		raw = append(raw, conf["Feature"]...)
+		raw = append(raw, conf["Features"]...)
+	}
+	return structutils.SortedDedup(raw)
+}
+
+func slurmdConfArgs(nodeset *slinkyv1beta1.NodeSet) []string {
+	confMap := map[string]string{"Features": strings.Join(baselineFeatures(nodeset), ",")}
+
+	// The NodeSet webhook validates ExtraConf at admission, but it is a separate,
+	// optional deployment, so the manager cannot assume it ran. On a parse error,
+	// degrade to the baseline (matching baselineFeatures) instead of failing the
+	// NodeSet's reconcile.
+	if conf, err := ParseExtraConf(nodeset.Spec.ExtraConf); err == nil {
+		for key, vals := range conf {
+			if key == "Feature" || key == "Features" {
+				continue
+			}
+			confMap[key] = strings.Join(vals, ",")
 		}
 	}
 
-	confList := []string{}
+	confList := make([]string, 0, len(confMap))
 	for key, val := range confMap {
 		confList = append(confList, fmt.Sprintf("%s=%s", key, val))
 	}
 	sort.Strings(confList)
-
-	args := []string{
-		"--conf",
-		fmt.Sprintf("'%s'", strings.Join(confList, " ")),
-	}
-
-	return args
+	return []string{"--conf", fmt.Sprintf("'%s'", strings.Join(confList, " "))}
 }
 
 func (b *WorkerBuilder) getWorkerHashes(ctx context.Context, nodeset *slinkyv1beta1.NodeSet) (map[string]string, error) {
--- a/internal/builder/workerbuilder/worker_app_test.go
+++ b/internal/builder/workerbuilder/worker_app_test.go
@@ -5,6 +5,7 @@ package workerbuilder
 
 import (
 	_ "embed"
+	"reflect"
 	"strings"
 	"testing"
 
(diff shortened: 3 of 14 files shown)


docs/usage/node-features.md
```diff
@@ -0,0 +1,106 @@
...
+apiVersion: v1
+kind: Node
+metadata:
+  name: node0
+  annotations:
+    features.slinky.slurm.net/spec: nn-75bfcf47ca3e4f7dc
+```
+
+The annotation is normally written by an external component that discovers each
+node's attributes (for example, a daemon querying a cloud provider's network
```
roma-glushko: This is great to mention how the annotations appear on the `Node` itself. Definitely was useful from my understanding.

Should we call them node features though here to be in line with K8s Node Feature Discovery naming?

```suggestion
node's features (for example, a daemon querying a cloud provider's network
```

internal/builder/workerbuilder/worker_app.go
```diff
@@ -283,53 +283,99 @@ func slurmdArgs(nodeset *slinkyv1beta1.NodeSet, controller *slinkyv1beta1.Contro
 	return args
 }
 
-func slurmdConfArgs(nodeset *slinkyv1beta1.NodeSet) []string {
-	extraConf := []string{}
-	if nodeset.Spec.ExtraConf != "" {
-		extraConf = strings.Split(nodeset.Spec.ExtraConf, " ")
+// ParseExtraConf parses a NodeSet's Spec.ExtraConf into a map of normalized keys
+// to their values. The expected shape is space-separated Key=Value pairs; each
```
roma-glushko, line 287: Are we pretty sure that comma separated key-value pairs would be enough here? I'm thinking about preserving a potential to evolve it from this state to anything slightly more complicated in terms of information passed inside without breaking changes. For example, could there be a case when we need to pass some additional params for some pairs?

Let me know if the question makes sense 👀

internal/builder/workerbuilder/worker_app.go
```diff
@@ -283,53 +283,99 @@ func slurmdArgs(nodeset *slinkyv1beta1.NodeSet, controller *slinkyv1beta1.Contro
...
+// value can be a comma-separated list (e.g. Features=a,b,c). Keys are title-cased
+// so "feature=a" and "Feature=a" both produce key "Feature".
+//
+// Tokens are split on whitespace, so leading, trailing, and repeated whitespace
+// are tolerated; any token without '=' returns an error. Exported so the NodeSet
+// validation webhook can reject malformed input with the same parser the worker
+// builder uses, keeping the accepted set identical at admission and at build time.
+func ParseExtraConf(extraConf string) (map[string][]string, error) {
+	out := map[string][]string{}
```
roma-glushko: It's nice to preallocate maps and slices for max performance:

```suggestion
	out := make(map[string][]string, 10)
```

I use some reasonable static number if I don't know the number of items in advance as in this case.

internal/builder/workerbuilder/worker_app_test.go
```diff
@@ -308,3 +309,192 @@ func TestWorkerBuilder_getResourceLimits(t *testing.T) {
...
+		{name: "repeated and surrounding whitespace tolerated", input: "  Weight=5  Feature=a ", want: map[string][]string{"Weight": {"5"}, "Feature": {"a"}}},
+		{name: "missing equals", input: "Weight10", wantErr: true},
+		{name: "missing equals among valid", input: "Weight=5 nope Feature=a", wantErr: true},
+		{name: "Feature and Features stay distinct keys", input: "Feature=a Features=b", want: map[string][]string{"Feature": {"a"}, "Features": {"b"}}},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			got, err := ParseExtraConf(tc.input)
+			if (err != nil) != tc.wantErr {
+				t.Fatalf("ParseExtraConf() err = %v, wantErr = %v", err, tc.wantErr)
```
roma-glushko, line 396: Doesn't seem like we are using [the testify/require library ](https://github.com/stretchr/testify) but we really should. It helps to reduce boilerplate coming from checks in tests.

Just a side note. Not directly related to this changeset.

internal/builder/workerbuilder/worker_app_test.go
```diff
@@ -308,3 +309,192 @@ func TestWorkerBuilder_getResourceLimits(t *testing.T) {
...
+		{name: "csv value splits", input: "Features=a,b,c", want: map[string][]string{"Features": {"a", "b", "c"}}},
+		{name: "case normalised", input: "feature=a", want: map[string][]string{"Feature": {"a"}}},
+		{name: "repeated and surrounding whitespace tolerated", input: "  Weight=5  Feature=a ", want: map[string][]string{"Weight": {"5"}, "Feature": {"a"}}},
+		{name: "missing equals", input: "Weight10", wantErr: true},
+		{name: "missing equals among valid", input: "Weight=5 nope Feature=a", wantErr: true},
+		{name: "Feature and Features stay distinct keys", input: "Feature=a Features=b", want: map[string][]string{"Feature": {"a"}, "Features": {"b"}}},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			got, err := ParseExtraConf(tc.input)
```
roma-glushko, line 394: These quick unit tests should be safe to run in parallel with `t.Parallel()`:


```suggestion
             t.Parallel()
			got, err := ParseExtraConf(tc.input)
```

I think it's actual to other tests too.

internal/controller/nodeset/slurmcontrol/slurmcontrol_test.go
```diff
@@ -405,6 +407,145 @@ func Test_realSlurmControl_UpdateNodeTopology(t *testing.T) {
 	}
 }
 
+func Test_realSlurmControl_UpdateNodeFeatures(t *testing.T) {
+	ctx := context.Background()
```
roma-glushko, line 416: I gotta double check out Golang version but we should be able to do this in the recent versions of the golang stdlib:

```suggestion
	ctx := t.Context()
```

internal/controller/nodeset/nodeset_sync.go
```diff
@@ -743,6 +750,77 @@ func (r *NodeSetReconciler) syncSlurmTopology(
...
+	}
+
+	return nil
+}
+
+// mergeFeatures returns the sorted, de-duplicated union of baseline and the
+// comma-separated features in annotation. It reuses workerbuilder.SortedDedup so
+// the baseline and the runtime merge share one normalization implementation.
+func mergeFeatures(baseline []string, annotation string) []string {
+	return workerbuilder.SortedDedup(append(append([]string{}, baseline...), strings.Split(annotation, ",")...))
```
roma-glushko: SortedDedup() func is one of those rare cases where we can place it in utils.go since it's used in a few places already and probably might be useful in other spots too.

SkylerMalinowski: We have a utility to merge arbitrary lists. Maybe this is better.

```go
	return workerbuilder.SortedDedup(structutils.MergeList(baseline, strings.Split(annotation, ",")))
```

roma-glushko (approved):
Looks great to me 👍

internal/builder/workerbuilder/worker_app.go
```diff
@@ -283,53 +283,99 @@ func slurmdArgs(nodeset *slinkyv1beta1.NodeSet, controller *slinkyv1beta1.Contro
...
+	for _, item := range strings.Fields(extraConf) {
```
SkylerMalinowski: gopls: Ranging over FieldsSeq is more efficient.

```go
	for item := range strings.FieldsSeq(extraConf) {
```

internal/builder/workerbuilder/worker_app.go
faganihajizada: Good question. There are two layers here::

1. **`ExtraConf` / `ParseExtraConf`** parses Slurm's `--conf` node-parameter syntax: space-separated `Key=Value`, the same syntax used for node lines in `slurm.conf`. `Spec.ExtraConf` is passed straight to `slurmd --conf`, so the format is fixed by Slurm, not chosen by us. See the [slurmd `--conf` docs](https://slurm.schedmd.com/slurmd.html) (e.g. `--conf "CPUs=16 RealMemory=30000 Gres=gpu:2"`).

2. **The annotation** is a comma-separated list of feature *names*, because that's exactly what a Slurm feature is. Per the [`slurm.conf` Features docs](https://slurm.schedmd.com/slurm.conf.html#OPT_Features): "A comma-delimited list of arbitrary strings ... there is no value or count associated with a feature ... a node either has a feature or it does not." Even a token that looks like `key=value` (e.g. `location=us-east-2`) is treated as one opaque string, which is why we comma-split only and pass tokens through verbatim.

On evolving it without breaking changes: if a future need is per-item types/counts/params, Slurm's own answer is GRES, not Features (the docs say "Also see Gres for being able to have more control such as types and count"). That would belong in a separate field/annotation for a different concept (GRES sync) rather than overloading this one.

internal/builder/workerbuilder/worker_app_test.go
faganihajizada: I agree. Safe in Go 1.26 and good practice, but the repo's existing table tests don't use it, so doing it only here is inconsistent. I will let maintainers decide here. cc @SkylerMalinowski

internal/controller/nodeset/slurmcontrol/slurmcontrol_test.go
faganihajizada: Available and slightly better than `context.Background()`, but again the repo uses `context.Background()` in tests today. maybe a separate PR to address this repo-wide. I will let @SkylerMalinowski to comment on this as well

internal/controller/nodeset/nodeset_sync.go
```diff
@@ -743,6 +750,77 @@ func (r *NodeSetReconciler) syncSlurmTopology(
...
+		}
+
+		annotation := node.Annotations[slinkyv1beta1.AnnotationNodeFeaturesSpec]
+		if annotation == "" {
+			// No features annotation: nothing to add beyond the baseline slurmd
+			// already registered with via --conf. Leave the node untouched so the
+			// operator only manages features on nodes that opt in via the
+			// annotation, rather than fighting an external node feature manager.
+			return nil
+		}
```
SkylerMalinowski: When a slurmd initially registers, its node_record_t is seeded with NodeSet extraConf (e.g. `--conf Feature(s)=<CSV>`). On successive registrations, `--conf` will be ignored, hence it will not change the node features automatically.

When `slurmControl.UpdateNodeFeatures()` is called, `slurmdConfArgs()` will consider both the Kube node annotation and the NodeSet `--conf Feature(s)` while determining the final features to update with. So if the Slurm node features differ, `slurmControl.UpdateNodeFeatures()` update to a known steady state. However, doing so may not be correct in all cases.

Aside from the registration and node record behavior, [NodeFeaturesPlugins](https://slurm.schedmd.com/slurm.conf.html#OPT_NodeFeaturesPlugins) can change the node features and active features.

I'm fine with ignoring the node feature re-registration quirk -- it's probably better to keep the NodeSet `--conf` Features in sync. However, supporting NodeFeaturesPlugins is a problem to untangle with respect to node (active) feature sync.

Rebooting a Slurm node is not supported, but [rebootless](https://slurm.schedmd.com/helpers.conf.html#OPT_rebootless) is a supported operation mode. So this PR should account for this operation mode and how node features can be updated out of the slurm-operator loop, but trying to support it may not be possible or at least not done easily.

This needs more consideration. Update on annotation introduces a bug (clobbering all nodes features when annotation is present); always updating forces idempotency but does not support NodeFeaturesPlugins in the current design.

internal/controller/nodeset/slurmcontrol/slurmcontrol.go
```diff
@@ -190,6 +192,55 @@ func (r *realSlurmControl) UpdateNodeTopology(ctx context.Context, nodeset *slin
...
+			return nil
+		}
+		return err
+	}
+
+	// Features are a set; slurmctld may return them in a different order than we
+	// sent, so compare order-insensitively to avoid issuing an update every reconcile.
+	desired := set.New(features...)
+	if desired.Equal(set.New(ptr.Deref(slurmNode.Features, slurmapi.V0044CsvString{})...)) &&
+		desired.Equal(set.New(ptr.Deref(slurmNode.ActiveFeatures, slurmapi.V0044CsvString{})...)) {
```
SkylerMalinowski: Use intermediate variables to make this easier to read and debug.

```go
	features := set.New(ptr.Deref(slurmNode.Features, slurmapi.V0044CsvString{})...)
	activeFeatures := set.New(ptr.Deref(slurmNode.ActiveFeatures, slurmapi.V0044CsvString{})...)
	if desired.Equal(features) && desired.Equal(activeFeatures) {
```

internal/builder/workerbuilder/worker_app_test.go
faganihajizada: Let's discuss with maintainers and if aligned, we can have a separate PR for repo-wide implementation. leaving out of scope of this PR

internal/controller/nodeset/nodeset_sync.go
faganihajizada: Good idea. https://github.com/SlinkyProject/slurm-operator/pull/203/changes/6c1c7d967a8197cb5768c7adb616c2e9c7e8e22e

SkylerMalinowski (commented):
The implementation assumes node features are idempotent from the slurm-operator's control. That is not always true -- consider [NodeFeaturesPlugins](https://slurm.schedmd.com/slurm.conf.html#OPT_NodeFeaturesPlugins), and an admin with `scontrol update nodename=.. features=... activefeatures=...`.

Some additional consideration is needed here.

internal/controller/nodeset/slurmcontrol/slurmcontrol_test.go
SkylerMalinowski: If the tests are stable, then I do not care.

Another PR should handle wholesale updates to use better testing patterns like `t.Context()` and `t.Parallel()`.

internal/builder/workerbuilder/worker_app_test.go
SkylerMalinowski: If the tests are stable, then I do not care.

Another PR should handle wholesale updates to use better testing patterns like `t.Context()` and `t.Parallel()`.

internal/builder/workerbuilder/worker_app_test.go
SkylerMalinowski: Another PR should handle better testing patterns like `t.Context()`, `t.Parallel()` and introduce `testify/require`.

SkylerMalinowski:
Please rebase and resolve conflicts. @faganihajizada

faganihajizada:
> Please rebase and resolve conflicts. @faganihajizada

@SkylerMalinowski done

SkylerMalinowski:
LGTM; I'll pull this in.

vivian-hafener:
Merged: https://github.com/SlinkyProject/slurm-operator/commit/43efef9655e0685a2119a7f44f30a12cf3bda977

PR #208, opened 2026-06-10
feat(controller): native Slurm active/passive HA via Replicas

afuller-crusoe:
## Summary

Add an optional Replicas field (default 1) to the Controller CRD. With replicas > 1 the operator enables Slurm native HA:

- one SlurmctldHost line per pod, addressed by per-pod DNS from the headless governing Service; SlurmctldTimeout=30 and MessageTimeout=5 defaults under HA (both user-overridable)
- required hostname anti-affinity so a primary and backup never share a node (user affinity overrides it)
- a shared RWX StateSaveLocation via persistence.existingClaim, enforced by CEL (per-pod volumeClaimTemplates are incoherent for HA)
- readiness gates on /livez instead of /readyz: a standby never answers /readyz 2xx, which would leave it permanently NotReady and wedge StatefulSet rolling updates; clients select the active controller via the SlurmctldHost list, not Service readiness
- sackd/slurmd receive every pod FQDN as a comma-separated --conf-server list and fall through past a standby
- logfile sidecar: never rm/recreate an existing FIFO and reopen after writer EOF; the old exit-and-recreate cycle raced a respawning slurmctld into a permanently blocked fifo_open on an unlinked inode

The governing Service keeps its pre-HA name in both modes because StatefulSet.spec.serviceName is immutable. It is headless (with publishNotReadyAddresses) only when replicas > 1; singleton controllers get the exact pre-HA Service, including user spec.service overrides (type, port, nodePort). Under HA those overrides are rejected instead of silently ignored: CEL covers service.port/nodePort at admission and the builder covers the schema-less spec wrapper at reconcile, stopping downstream sync steps so config and StatefulSet are never built against a Service of the wrong shape. Because clusterIP is also immutable, the sync deletes and recreates an operator-owned Service when toggling replicas across 1 changes headlessness, emitting an event; per-pod DNS therefore works when existing clusters scale into HA.

Singleton behavior (replicas 1 or unset) is unchanged. Validated on a live cluster: takeover in SlurmctldTimeout, jobs submitted/completed by the backup during a 5-minute primary outage, rolling updates proceed through the standby.

## Checklist

- [x] I have read
  and the

## Breaking Changes

None for existing configurations: with `replicas: 1` (or unset), the
controller Service, probes, and `slurm.conf` are identical to today, including
`service` type/port/nodePort overrides.

With `replicas > 1` (opt-in), controller behavior changes by design:

- The controller Service becomes headless to provide the per-pod DNS Slurm HA
  requires; since `clusterIP` is immutable, the operator deletes and recreates
  it when `replicas` crosses 1 (brief blip, event emitted).
- `service` overrides are rejected (CRD validation + reconcile error) instead
  of applied; external exposure of an HA pair is out of scope.
- `slurm.conf` gets one `SlurmctldHost` per pod plus `SlurmctldTimeout=30` /
  `MessageTimeout=5` defaults (overridable via `extraConf`).
- Readiness gates on `/livez` (a standby never answers `/readyz`, which would
  wedge rolling updates), pods get required hostname anti-affinity, and
  `persistence.existingClaim` must be a ReadWriteMany PVC.

## Testing Notes

**Failover (live cluster, Slurm 25.11.2):** killing the primary never
interrupts running jobs; control-plane loss is bounded by `SlurmctldTimeout`
(30 s), with the backup accepting and completing jobs through a 5-minute
primary outage. In practice the StatefulSet recreated the primary in ~25 s and
it resumed from shared statesave (`srun` verified immediately after).

Also validated live: 2→1→2 scaling recreated the Service with the right shape
each time (ClusterIP ↔ headless, event emitted, `slurm.conf` regenerated); a
singleton honored a `NodePort` override; admission rejected `nodePort` +
`replicas: 2`.

`make test` covers the rest: singleton Service is byte-for-byte pre-HA, HA
Service is headless with overrides rejected, Service recreation in both toggle
directions, and no deletion of Services the operator doesn't own.

## Additional Context

https://slurm.schedmd.com/quickstart_admin.html#HA


SkylerMalinowski:
Hi @afuller-crusoe ,

Thanks for the PR and your interest in Slurm HA. Before the implementation review of this PR, we need to discuss why this feature/functionality is needed. What does Slurm HA solve that Kubernetes pod regeneration does not? Is the baggage of Slurm HA worth it?

afuller-crusoe:
Kubernetes will restart pods which fail liveness probes or crash, if this is a short temporary blip then pod restarting is fast and  kubernetes can have minimal downtime, even compared to slurm HA. However, if this is not a temporary blip then Slurm HA is necessary to ensure high uptime for the controller. Two examples of how relying only on k8s fails:

* Controller pod failing liveness probes due to localized networking issues: Here the pod may be SIGTERM'ed by the kubelet, but is simply restarted and can enter CrashLoopBackoff. **The pod is not rescheduled** in this case resulting in significant downtime and requires manual intervention to get the pod rescheduled
* Node moves to not ready state/loses connection to API server: Here the controller can be down and not serving the slurm cluster since its on a bad node. Also results in significant downtime, however in this case the pod will eventually be terminated and rescheduled to a good node. 

Slurm HA minimizes the impact of these long-running failures by providing an actual backup which can start serving slurmctld quickly (~30 seconds) until the underlying condition causing node/localized pod failures are resolved. I have specifically included anti-affinity to make sure all replicas end up on different nodes.

Given we have seen these sorts of issues in our own cloud, and resulted in significant downtime until manual remediation, the baggage that comes with slurm HA is worth it. To track the baggage here is what I see, please let me know if there is more I am not tracking:
* Requires shared storage between all slurmctld replicas (shared across nodes)
* Breaks nodeport/clusterIP service for multi-replicas since it would route to the primary and standy backups (which cannot serve).

internal/builder/controllerbuilder/controller_config.go
```diff
@@ -204,6 +221,14 @@ func buildSlurmConf(
...
 	conf.AddProperty(config.NewProperty("MaxNodeCount", 1024))
 	conf.AddProperty(config.NewProperty("GresTypes", "gpu"))
+	if replicas > 1 {
+		// Emitted before ExtraConf so users may override. SlurmctldTimeout=30
+		// suits in-cluster heartbeat latency; MessageTimeout=5 turns the
+		// silent ~10s per-RPC hang against a dead primary into a fast
+		// timeout-then-retry.
+		conf.AddProperty(config.NewProperty("SlurmctldTimeout", 30))
+		conf.AddProperty(config.NewProperty("MessageTimeout", 5))
+	}
```
SkylerMalinowski, line 231: The operator should not opinionated unless it is required configuration. This does not look required, so it should not exist.

Currently the operator does set non-required defaults, but those will be going away in v1.2.

api/v1beta1/controller_keys.go
```diff
@@ -118,3 +118,24 @@ func (o *Controller) ConfigKey() types.NamespacedName {
...
 	}
 }
+
+// Replicas returns the configured slurmctld replica count, defaulting to 1.
+// External controllers are always treated as a single replica.
+func (o *Controller) Replicas() int32 {
+	if o.Spec.External {
+		return 1
+	}
+	return ptr.Deref(o.Spec.Replicas, 1)
```
SkylerMalinowski, line 128: Should codify with the `defaults` module and use that const for non-default Go values. See the `defaults` module for details.

api/v1beta1/controller_types.go
```diff
@@ -21,13 +21,26 @@ var (
...
+	// Replicas is the number of slurmctld pods to run. 1 (the default) is a
+	// singleton restarted by Kubernetes on failure. Greater than 1 enables
+	// Slurm native active/passive HA: a primary plus backups spread across
+	// nodes, with automatic failover after SlurmctldTimeout. Requires
+	// persistence.existingClaim to reference a ReadWriteMany PVC for
+	// StateSaveLocation.
+	// Ref: https://slurm.schedmd.com/quickstart_admin.html#HA
+	// +kubebuilder:validation:Minimum=1
+	// +optional
+	Replicas *int32 `json:"replicas,omitempty"`
```
SkylerMalinowski, line 42: Creating implicit functionality by overloading the replicas concept is not great. I can see why you did so, but promoting HA into an object would probably be best.

Something like the following:

```yaml
controller:
  ha:
    enabled: true
    backups: 2  
```

SkylerMalinowski (commented):
You point out good edge cases which Slurm HA should help solve. Thank you for providing use cases for this functionality.

Looking over the design, there is room for improvement. I have left you some feedback so you are not left in the dark while I take it internally and work on it.

You do not have to act on the comments, I will handle them internally but I'm open to suggestions and feedback.

internal/builder/common/scripts/logfile.sh
```diff
@@ -7,10 +7,15 @@ set -eu
...
 	mkfifo -m 777 "$SOCKET"
 fi
-while IFS="" read data; do
-	echo "$data"
-done <"$SOCKET"
+# Reopen after every writer close (EOF); exiting would restart the container,
+# recreate the FIFO, and re-expose the unlinked-inode race.
+while true; do
+	cat "$SOCKET" || true
+done
```
SkylerMalinowski, line 21: For future reference, putting unrelated changes and fixes into a feature commit is not great.

Under what conditions were you seeing the logfile sidecar hit EOF and restart?

afuller-crusoe: Yes, its not directly related, since this bug hits in the singleton case. I was seeing this occur often in my testing since I was killing the slurmctld process, and this bug manifest often, therefore included. 

Every time the slurmctld process exited logfile would hit EOF and restart. This then created a race where slurmctld on restart would open the socket, before it was recreated. Since its a race it didn't always happen. Given this failure mode can exist in singleton mode I think its fine to remove from this commit, but should still be tracked

SkylerMalinowski:
I've been working on getting this prepared for internal review. I just wanted to let you know that the shape of the final commit attributed you will be slightly different in design. Thank you for the initial work on this.

Of the high level design changes (pending):

- always back the Controller StatefulSet with a headless service (in addition to the customizable one)
- restructure CRD expression of HA configuration into a sub-object (similar to the one in review comments)
- add more robust validation of existingClaim compatibility
- allow post deployment enable/disable of HA as long as prerequisites are met (e.g. PVC with ReadWriteMany)
- remove forced slurm.conf settings (as commented in review)

No need for you to take further action on this PR. Feedback is always welcome.

internal/builder/common/scripts/logfile.sh
SkylerMalinowski: The logfile changes were sliced off and already merged in 4abbf5395c286159fdab6e1aab3fc58be904a4b0 (merge commit 9c967b0b19bba88c4eb731df77b9c3f2987ac3b2).

afuller-crusoe:
Hey Skyler, any updates here? Would you like me to implement the changes called out in this review?

SkylerMalinowski:
While it is fairly straight forward to add this functionality, internally we are discussing how Slurm HA aligns with Kubernetes and potential future Slurm plans. I already have converted this PR into an internal MR which is getting internal review and discussion.

What would be helpful is some more information. Are there any other cases where Slurm HA would solve edge cases where relying on pod regeneration is not sufficient?

Drilling further into your previous statements:

> * Controller pod failing liveness probes due to localized networking issues: Here the pod may be SIGTERM'ed by the kubelet, but is simply restarted and can enter CrashLoopBackoff. **The pod is not rescheduled** in this case resulting in significant downtime and requires manual intervention to get the pod rescheduled

Is this failure mode due to slurm.conf drift due to implicit bad reconfigure recovery in Slurm? For example, if you apply an invalid slurm.conf after be initially valid, when slurmctld reconfigures the new config is invalid and the reconfigure is aborted, thus the old slurm.conf is still being used but the Kubernetes API contains the invalid config. If the slurmctld pod is restarted, it will fail because there is no locally cached valid slurm.conf.

In Slinky v1.2, you can set `inplaceReconfigure=false` to make slurm.conf updates always force a slurmctld pod restart, ensuring that the Kubernetes API the source of truth.

PR #214, opened 2026-06-16
helm: fix vendor.dcgm.enabled: false has no effect

### LINKED ISSUE #210: [BUG]: vendor.nvidia.dcgm.enabled: false has no effect — DCGM prolog always wired into prologScriptRefs
### Version

1.1.0

### Which installation method(s) does this occur on?

Helm / Kubernetes manifests

### Describe the bug.

Setting `vendor.nvidia.dcgm.enabled: false` (the default) does not prevent the DCGM prolog/epilog scripts from being added to `prologScriptRefs`/`epilogScriptRefs` in the Controller CR. As a result, the Slinky operator always injects `Prolog=prolog-90-dcgm.sh` and `PrologFlags=Contain` into `slurm.conf`, even on CPU-only clusters where DCGM is not intended to be used. This prevents workers from starting successfully on some CPU environments (minikube on macOS here).

### Minimum reproducible example

```shell
# Render the chart with dcgm explicitly disabled (the default)
helm template my-cluster oci://ghcr.io/slinkyproject/charts/slurm \
  --version 1.1.0 \
  --set vendor.nvidia.dcgm.enabled=false \
  | grep -E "prologScriptRefs|epilogScriptRefs|dcgm"

# Observed output — DCGM refs are present despite enabled=false:
#   prologScriptRefs:
#     - name: my-cluster-slurm-prolog-dcgm
#   epilogScriptRefs:
#     - name: my-cluster-slurm-epilog-dcgm

# Also confirm with enabled= (empty string, the only working workaround):
helm template my-cluster oci://ghcr.io/slinkyproject/charts/slurm \
  --version 1.1.0 \
  --set "vendor.nvidia.dcgm.enabled=" \
  | grep -E "prologScriptRefs|epilogScriptRefs|dcgm"

# With empty string the refs are correctly absent.
```

### Relevant log output

```shell
[2026-06-15T23:55:53] error: prolog failed: rc:127 output:[2026-06-15T23:55:53] error: _run_command_child_exec: execv(/usr/sbin/slurmd): No such file or directory
```

### Additional environment details

```shell
Helm: 3.x
Kubernetes: minikube v1.35 (single-node, cgroup v2, IgnoreSystemd=yes)
slurm chart: 1.1.0
Slinky operator: 1.1.0
Slurm: 26.05.0
```

### Other/Misc.

_No response_

### Code of Conduct

- [x] I agree to follow Slurm Operator's Code of Conduct
- [x] I have searched the [open bugs](https://github.com/SlinkyProject/slurm-operator/issues?q=is%3Aopen+is%3Aissue+label%3Abug) and have found no duplicates for this bug report

shravyavorugallu:
Fixes #210

## Root Cause

The `vendor.dcgm.enabled` helper in `_helpers.tpl` used Sprig's `ternary` function via a pipeline:

```go-template
{{- ($vendor | dig "nvidia" "dcgm" "enabled" false) | ternary "true" "" -}}
```

When `vendor.nvidia.dcgm.enabled: false` is set, `dig` returns a Go `bool(false)`. Sprig's `ternary` reflects this value and can evaluate it as truthy, causing DCGM prolog/epilog ConfigMap refs to always be injected into the Controller CR.

This breaks CPU-only deployments (minikube, non-GPU nodes) because the operator tries to reconcile DCGM prolog scripts that were never created.

## Fix

Replace `ternary` with a native Go template `if` block, which correctly evaluates boolean `false` as falsy in all Helm versions:

```go-template
{{- if ($vendor | dig "nvidia" "dcgm" "enabled" false) -}}
true
{{- end -}}
```

The rendered output is unchanged: `"true"` when enabled, `""` when disabled.

### DIFF
--- a/helm/slurm/templates/vendor/nvidia/dcgm/_helpers.tpl
+++ b/helm/slurm/templates/vendor/nvidia/dcgm/_helpers.tpl
@@ -9,7 +9,9 @@ Check if DCGM integration is enabled
 */}}
 {{- define "vendor.dcgm.enabled" -}}
 {{- $vendor := .Values.vendor | default dict -}}
-{{- ($vendor | dig "nvidia" "dcgm" "enabled" false) | ternary "true" "" -}}
+{{- if ($vendor | dig "nvidia" "dcgm" "enabled" false) -}}
+true
+{{- end -}}
 {{- end }}
 
 {{/*


vivian-hafener:
Good morning,

Your issue indicates that this bug is present on `v1.1.0`, but the root cause you present as problematic is only present on `main`. On `v1.1.0`, the logic is [as follows](https://github.com/SlinkyProject/slurm-operator/blob/v1.1.0/helm/slurm/templates/vendor/nvidia/dcgm/_helpers.tpl#L7C1-L12C11):
```tpl
{{/*
Check if DCGM integration is enabled
*/}}
{{- define "vendor.dcgm.enabled" -}}
{{- .Values.vendor.nvidia.dcgm.enabled -}}
{{- end }}
```
The logic in `v1.1.0` was broken, which did result in `vendor.dcgm.enabled: false` having no effect. However, I believe that this was resolved by https://github.com/SlinkyProject/slurm-operator/commit/f1786e45a8337a2d4522a5315de4d18b6c6e2a0c on `main`.

If you can confirm that that commit does resolve the behavior that you are seeing, I can look into backporting that commit to `release-1.0` and `release-1.1`.

Best,
Vivian Hafener

shravyavorugallu:
Hi Vivian,
Thanks for the clarification, that helps.

I can confirm that the behavior I’m seeing does reproduce on v1.1.0, where
setting "vendor.dcgm.enabled: false" has no effect due to the logic you
pointed out.

I’ve also reviewed commit "f1786e4" on "main", and it does appear to
address the issue. Based on that, a backport to
"release-1.1" would be great.

Thankyou,
Shravya Vorugallu.

vivian-hafener:
Good afternoon @shravyavorugallu,

This was cherry-picked to `release-1.1` and is present in `v1.1.1`: b4ade447411eae10c90d665047d6dd21fc6f2699

Upon further review, this fix is not needed for `release-1.0`.

Best,
Vivian Hafener

PR #217, opened 2026-06-24
Expose validating webhook namespaceSelector

faganihajizada:
## Summary

Expose namespace scoping for the operator's admission webhooks and make the default exclusions safe.

- Add `webhook.validating.namespaceSelector`: when set it is rendered verbatim on the six validating (CRD) webhooks; when unset the existing computed selector (the `webhook.namespaces` In-list plus the system-namespace NotIn exclusions) is used. Follows cert-manager's pass-through `namespaceSelector` pattern on its stateless CRD webhooks.
- Add `kube-node-lease` to the default `NotIn` exclusions for every webhook, per the [Kubernetes admission webhook good practices](https://kubernetes.io/docs/concepts/cluster-administration/admission-webhooks-good-practices/)
- The `pods/binding` mutating webhook is intentionally not given an override: its handler reads the bound Pod from a namespace-scoped cache (`--namespaces` -> controller-runtime `Cache.DefaultNamespaces`), so its scope stays single-sourced via `webhook.namespaces` (which already drives both the cache and the selector)
- Refactor the seven duplicated inline `namespaceSelector` blocks into a `slurm-operator.webhook.namespaceSelector` helper.

## Checklist

- [x] I have read CONTRIBUTING.md and the Code of Conduct.

## Breaking Changes

None. Behavior is unchanged except that every webhook now also excludes the `kube-node-lease` namespace (an additive safety exclusion; no Slurm-managed resources live there). `webhook.validating.namespaceSelector` defaults to `{}`, so there is no change unless an operator sets it.

## Testing Notes

- `helm unittest --strict` on the `slurm-operator` chart: 104 pass (12 snapshots), including new cases asserting the validating override renders verbatim and does not leak into the `pods/binding` webhook.
- `helm lint --strict`: passes.
- Regenerated `tests/__snapshot__/webhook_webhook_test.yaml.snap` and the chart `README.md` (`make helm-docs`); the snapshot diff is only the added `kube-node-lease` lines.
- Live-verified on a kind cluster (K8s v1.35) by applying the rendered webhook configurations to a real API server and reading them back:
  - default: validating `NotIn [kube-system, kube-node-lease]`; binding `NotIn [kube-system, kube-node-lease, <operator-ns>]`
  - `webhook.validating.namespaceSelector` set: validating renders the override verbatim; binding unchanged
  - `webhook.namespaces=slurm,compute`: both add `In [slurm, compute]`

## Additional Context

`objectSelector` is deliberately not exposed

### DIFF
--- a/helm/slurm-operator/README.md
+++ b/helm/slurm-operator/README.md
@@ -99,4 +99,5 @@ Kubernetes: `>= 1.29.0-0`
 | webhook.topologySpreadConstraints | list | `[]` | Topology spread constraints for pod assignment. Prefer scheduling replicas across failure domains (nodes, zones, ...) when running in HA. Ref: https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/ |
 | webhook.validating.failurePolicy | string | `"Fail"` | Action taken when the validating admission webhook is unreachable or returns an error. Ref: https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#failure-policy |
 | webhook.validating.matchPolicy | string | `"Equivalent"` | How the rules listed in the validating webhook are matched against incoming requests. Ref: https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-matchpolicy |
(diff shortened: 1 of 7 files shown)


vivian-hafener:
Merged: https://github.com/SlinkyProject/slurm-operator/commit/f48b5c0ca9056740a5b28c5790aafc1db4187dc1

PR #218, opened 2026-06-26
fix(webhook): reuse Helm-native serving cert on upgrade

giuliocalzo:
## Summary

When `certManager.enabled=false`, the operator chart generates the webhook serving certificate in Helm using Sprig's `genCA`/`genSignedCert`. Because those functions produce fresh material on every render, each `helm upgrade` rotated the CA, the serving cert, and the injected `caBundle`. The running webhook pod is not restarted on upgrade (no checksum annotation), so it kept serving the **old** cert while the API server was updated to trust only the **new** CA — breaking admission until the pod was manually rolled.

This PR makes the Helm-native path reuse the existing certificate:

- `lookup` the existing TLS `Secret`; if it has a complete cert (`tls.crt`/`tls.key`/`ca.crt`), reuse it verbatim for both the `Secret` and the injected `caBundle`.
- Regenerate only when the `Secret` is missing/incomplete — i.e. first install, or `helm template`/`--dry-run` where `lookup` returns an empty map.
- The CA stored in the `Secret` and the CA injected as `caBundle` come from the same source, so they stay consistent within a render.

cert-manager mode (`certManager.enabled=true`, the default) is unchanged.

## Checklist

## Breaking Changes

None. cert-manager mode (the default) is unchanged, and the Helm-native path now preserves existing certificate material across upgrades instead of rotating it.

## Testing Notes

- `helm template` with `certManager.enabled=false`: single TLS `Secret`, one unique `caBundle`, `Secret.ca.crt == caBundle`.
- `helm template` with `certManager.enabled=true`: no native `Secret`/`caBundle`, cert-manager `Issuer`/`Certificate` + inject annotations present (default snapshot unchanged).
- `helm unittest ./helm/slurm-operator`: 105 tests / 12 snapshots pass, including:
  - generation on first install (base64 PEM material, `caBundle` on every webhook, no cert-manager annotations),
  - verbatim reuse of an existing `Secret` mocked via `kubernetesProvider`,
  - cert-manager-enabled path renders no native `Secret`/`caBundle`.
- `helm lint` passes with cert-manager both enabled and disabled.

## Additional Context

Fixes broken admission after `helm upgrade` in the Helm-native (non cert-manager) certificate mode.

### DIFF
--- a/helm/slurm-operator/templates/webhook/webhook.yaml
+++ b/helm/slurm-operator/templates/webhook/webhook.yaml
@@ -5,25 +5,46 @@ SPDX-License-Identifier: Apache-2.0
 
 {{- if .Values.webhook.enabled }}
 {{- $certNamespacedName := printf "%s/%s" (include "slurm-operator.namespace" .) .Values.certManager.secretName -}}
-{{- $durationDays := int 3650 -}}
-{{- $ca := genCA .Values.certManager.secretName $durationDays -}}
+{{- $caBundle := "" -}}
 {{- if not .Values.certManager.enabled }}
-{{- $cn := (include "slurm-operator.webhook.name" .) -}}
-{{- $dnsList := list (printf "%s.%s" (include "slurm-operator.webhook.name" .) (include "slurm-operator.namespace" .)) (printf "%s.%s.svc" (include "slurm-operator.webhook.name" .) (include "slurm-operator.namespace" .)) -}}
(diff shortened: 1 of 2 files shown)


vivian-hafener:
Good afternoon @giuliocalzo,

I will look into this.

Best,
Vivian Hafener

giuliocalzo:
merged in https://github.com/SlinkyProject/slurm-operator/commit/089deb2467c68be42e95564941e7a20c92da6099

PR #221, opened 2026-07-02
fix(controllers): return non-nil list from refResolver on List error

faganihajizada:
## Summary

The four `refResolver.Get*ForController` / `GetControllersForAccounting` list helpers (`internal/utils/refresolver/refresolver.go`) returned `(nil, err)` when the underlying `List` failed, even though they return a non-nil empty list for their nil-argument guard branch. Several event handlers log that error and then range over the returned list to enqueue work, so on a transient List failure (cache not yet synced at startup, or an apiserver blip) they dereferenced the nil list and panicked in the informer event path. Affected handlers: the NodeSet, RestApi, and LoginSet Secret event handlers, and the RestApi controller event handler.

This fixes the problem at the source: all four resolver methods now return a non-nil empty list alongside the error, matching their own nil-argument branch, so every caller is panic-safe.

## Checklist

- [x] Documentation is updated if user-visible behavior changes

## Breaking Changes

N/A

## Testing Notes

Unit: the List-error case is folded into each of the four per-method resolver tests as a "list error" row that injects a failing `List` via `WithInterceptorFuncs`, asserting a non-nil list is returned with the error. These rows fail against the previous implementation (the returned list is nil) and pass with the fix. `go test ./internal/utils/refresolver/... ./internal/controller/{nodeset,restapi,loginset}/eventhandler/...` and `golangci-lint run` are clean.

Live (kind): built the operator from this branch, deployed it over a running Slurm cluster, and rotated the referenced auth secret. The Secret event handler correctly enqueued and reconciled the RestApi, Controller, and NodeSet, with zero panics and no operator restarts. The error path itself (a transient cache List failure) is not reliably reproducible on a live cluster, so the unit test is the deterministic proof; the live run confirms no regression on the normal enqueue path.

### DIFF
--- a/internal/utils/refresolver/refresolver.go
+++ b/internal/utils/refresolver/refresolver.go
@@ -56,7 +56,7 @@ func (r *RefResolver) GetNodeSetsForController(ctx context.Context, controller *
 
 	list := &slinkyv1beta1.NodeSetList{}
 	if err := r.reader.List(ctx, list, client.InNamespace(controller.Namespace)); err != nil {
-		return nil, err
+		return &slinkyv1beta1.NodeSetList{}, err
 	}
 
 	out := &slinkyv1beta1.NodeSetList{}
@@ -80,7 +80,7 @@ func (r *RefResolver) GetLoginSetsForController(ctx context.Context, controller 
 
 	list := &slinkyv1beta1.LoginSetList{}
 	if err := r.reader.List(ctx, list, client.InNamespace(controller.Namespace)); err != nil {
-		return nil, err
+		return &slinkyv1beta1.LoginSetList{}, err
 	}
 
 	out := &slinkyv1beta1.LoginSetList{}
@@ -104,7 +104,7 @@ func (r *RefResolver) GetRestapisForController(ctx context.Context, controller *
 
 	list := &slinkyv1beta1.RestApiList{}
 	if err := r.reader.List(ctx, list, client.InNamespace(controller.Namespace)); err != nil {
-		return nil, err
+		return &slinkyv1beta1.RestApiList{}, err
 	}
 
 	out := &slinkyv1beta1.RestApiList{}
@@ -128,7 +128,7 @@ func (r *RefResolver) GetControllersForAccounting(ctx context.Context, accountin
 
 	list := &slinkyv1beta1.ControllerList{}
 	if err := r.reader.List(ctx, list, client.InNamespace(accounting.Namespace)); err != nil {
-		return nil, err
+		return &slinkyv1beta1.ControllerList{}, err
 	}
 
 	out := &slinkyv1beta1.ControllerList{}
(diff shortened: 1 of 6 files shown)


internal/utils/refresolver/refresolver.go
```diff
@@ -56,7 +56,7 @@ func (r *RefResolver) GetNodeSetsForController(ctx context.Context, controller *
 
 	list := &slinkyv1beta1.NodeSetList{}
 	if err := r.reader.List(ctx, list, client.InNamespace(controller.Namespace)); err != nil {
-		return nil, err
+		return &slinkyv1beta1.NodeSetList{}, err
```
SkylerMalinowski: It best practice in Golang to return the empty value when an error occurs, hence this is bad form.

internal/utils/refresolver/refresolver.go
```diff
@@ -80,7 +80,7 @@ func (r *RefResolver) GetLoginSetsForController(ctx context.Context, controller
 
 	list := &slinkyv1beta1.LoginSetList{}
 	if err := r.reader.List(ctx, list, client.InNamespace(controller.Namespace)); err != nil {
-		return nil, err
+		return &slinkyv1beta1.LoginSetList{}, err
```
SkylerMalinowski: It best practice in Golang to return the empty value when an error occurs, hence this is bad form.

internal/utils/refresolver/refresolver.go
```diff
@@ -104,7 +104,7 @@ func (r *RefResolver) GetRestapisForController(ctx context.Context, controller *
 
 	list := &slinkyv1beta1.RestApiList{}
 	if err := r.reader.List(ctx, list, client.InNamespace(controller.Namespace)); err != nil {
-		return nil, err
+		return &slinkyv1beta1.RestApiList{}, err
```
SkylerMalinowski: It best practice in Golang to return the empty value when an error occurs, hence this is bad form.

internal/utils/refresolver/refresolver.go
```diff
@@ -128,7 +128,7 @@ func (r *RefResolver) GetControllersForAccounting(ctx context.Context, accountin
 
 	list := &slinkyv1beta1.ControllerList{}
 	if err := r.reader.List(ctx, list, client.InNamespace(accounting.Namespace)); err != nil {
-		return nil, err
+		return &slinkyv1beta1.ControllerList{}, err
```
SkylerMalinowski: It best practice in Golang to return the empty value when an error occurs, hence this is bad form.

SkylerMalinowski (commented):
Aside from the comments, everything else is fine.

faganihajizada:
> Aside from the comments, everything else is fine.

Thanks for review @SkylerMalinowski! All addressed

vivian-hafener:
Merged: https://github.com/SlinkyProject/slurm-operator/commit/5a8a6867f55ed319ff6951087feccb268352a60e and https://github.com/SlinkyProject/slurm-operator/commit/3271cdeefcf91fed887954db7533c4a8b033af51

PR #225, opened 2026-07-03
fix: wire the metrics-secure flag to secure the metrics endpoint

faganihajizada:
## Summary

Both the operator and webhook managers parse a `--metrics-secure` flag whose help text says "If set the metrics endpoint is served securely", but the flag was never applied: the metrics server options only set `BindAddress` and `TLSOpts`, so metrics were always served as plain unauthenticated HTTP regardless of the flag.

This wires the flag end to end, following the standard kubebuilder scaffold pattern and the controller-runtime documentation.

While wiring this, the chart's metrics port handling was hardened with a validating helper: `metricsPort` must be a decimal integer between 0 and 65535 (typos, leading-zero strings, out-of-range and overflowing values now fail at render time instead of silently disabling metrics or crashing at runtime), the chart now passes `--metrics-addr "0"` explicitly when the port is 0 so that the documented "0 disables it" contract actually holds for the operator (whose binary default is `:8080`), and rendering fails fast when `metricsSecure` is enabled without a non-zero port.

## Checklist

## Breaking Changes

N/A
the flag and both chart values default to off, and a default install renders the same operator args as before. The webhook's default rendered args now include an explicit `--metrics-addr "0"`, which is a no-op because the webhook binary's default is already "0".

## Testing Notes

`helm-unittest` suites cover the new behavior (secure flag on and off for both components, explicit disable adjacency by argument index, a positive custom-port case, render failures for secure-without-port, string "0", non-numeric, leading-zero and out-of-range ports, and metrics-reader ClusterRole gating, naming and rules); the full chart suite passes.

Live (kind): deployed the operator and webhook built from this branch. With default values, plain HTTP scraping still returns 200 with no token (backward compatible). With `metricsSecure` enabled (operator :8080, webhook :8443): requests without a token get 401; a ServiceAccount token bound to the shipped metrics-reader ClusterRole gets 200 with the metrics body, which exercises the TokenReview and SubjectAccessReview RBAC end to end; an authenticated but unbound ServiceAccount gets 403; plain HTTP against the TLS endpoint is rejected. A `helm upgrade` with an invalid port was rejected at render time by the new guard. No panics or restarts.


vivian-hafener:
Merged: https://github.com/SlinkyProject/slurm-operator/commit/a53eac6433eb06bebbcef6bfb493d3f2bc1bb51e

PR #226, opened 2026-07-03
fix(nodeset): sanitize Slurm drain reason from Node annotation

giuliocalzo:
## Summary

The `AnnotationNodeCordonReason` value on a Kubernetes Node is user-controlled and was used verbatim as the Slurm node `Reason` in `internal/controller/nodeset/nodeset_sync.go`. Anyone able to annotate a Node could inject arbitrary or control characters (e.g. newlines) into the Slurm node Reason, enabling log/reason spoofing.

This sanitizes the annotation before use via a new `sanitizeSlurmReason` helper, which strips control characters, replaces newlines/tabs with spaces, trims whitespace, and truncates to a bounded, rune-safe length.

## Checklist

- [x] I have read [CONTRIBUTING.md](https://github.com/SlinkyProject/slurm-operator/blob/main/CONTRIBUTING.md)  and the  [Code of Conduct](https://github.com/SlinkyProject/slurm-operator/blob/main/CODE_OF_CONDUCT.md).

## Breaking Changes

None. The sanitization only affects malformed/malicious reason strings; well-formed reasons are unchanged.

## Testing Notes

`go test ./internal/controller/nodeset/ -run Test_sanitizeSlurmReason`

Adds table-driven tests covering the sanitization behavior (control characters, newlines/tabs, trimming, and length truncation).

## Additional Context

Hardening fix against log/reason spoofing via user-controlled Node annotations.

### DIFF
--- a/internal/controller/nodeset/nodeset_sync.go
+++ b/internal/controller/nodeset/nodeset_sync.go
@@ -10,6 +10,7 @@ import (
 	"sort"
 	"strings"
 	"time"
+	"unicode"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


vivian-hafener:
Merged: https://github.com/SlinkyProject/slurm-operator/commit/5de5c5eecfbe43e08f960e143bb172a677242154

PR #229, opened 2026-07-06
fix(nodeset): count unavailable pods against rolling update maxUnavailable

faganihajizada:
## Summary

During a RollingUpdate, `syncRollingUpdate` passed only the healthy old-revision pods to `splitUpdatePods`, so the `newPods` split inside it was always empty and `numUnavailable` was always zero. Every reconcile therefore condemned a fresh `maxUnavailable`-sized batch of old pods regardless of whether any replacement was ready, and a slow or broken image pull could drain a whole NodeSet to zero available workers. 

Reproduced on a kind cluster with `replicas=3` and `maxUnavailable=1` and an unpullable image: the operator logged 43 condemnation events in under three minutes and Slurm capacity dropped to 1 of 3 nodes in service where the budget guarantees 2, with visible cordon/uncordon flapping because `doPodProcessing` (which receives the full pod list) and `syncRollingUpdate` disagreed about the same pods.

Both callers now pass the full pod list and the accounting lives in `splitUpdatePods`: unhealthy old pods keep their replica slot without consuming the budget and are excluded from the deletion candidates. Daemonset mode keeps the previous seed because its pod set is not bounded by `Replicas`, so terminating remnants on nodes that left the eligible set must not stall the update.

## Checklist

## Breaking Changes

N/A

## Testing Notes

New unit tests cover the throttle (an unavailable new pod blocks further condemnation), the 8281235 exemption in both directions (an unhealthy old pod is deleted while the budget is exhausted, and it does not consume budget that a healthy old pod needs), terminating pods, scale-in remnants, and percentage budgets, with cordon and deletion assertions against the fake client, and the pre-existing update case now asserts that a ready new pod lets the next old pod be condemned; three of the new cases fail against the previous code.

Live on kind with the fixed operator (`replicas=3`, `maxUnavailable=1`): an update to an unpullable image condemned exactly one pod, the wedged replacement stayed the only unavailability, and the two remaining workers stayed Ready and idle in Slurm with no further condemnations while the pull kept failing. A subsequent valid update combined with a concurrent scale-down from 3 to 1 replicas converged to a single updated Ready pod, with the last old pod condemned for update while a scale-in remnant was still terminating, confirming remnants do not consume the budget.

### DIFF
--- a/internal/controller/nodeset/nodeset_sync.go
+++ b/internal/controller/nodeset/nodeset_sync.go
@@ -1527,7 +1527,7 @@ func (r *NodeSetReconciler) syncRollingUpdate(
 
 	_, oldPods := findUpdatedPods(pods, hash)
 
-	unhealthyPods, healthyPods := nodesetutils.SplitUnhealthyPods(oldPods)
+	unhealthyPods, _ := nodesetutils.SplitUnhealthyPods(oldPods)
 	if len(unhealthyPods) > 0 {
 		logger.Info("Delete unhealthy pods for Rolling Update",
 			"unhealthyPods", len(unhealthyPods))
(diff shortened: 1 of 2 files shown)


vivian-hafener:
Merged: https://github.com/SlinkyProject/slurm-operator/commit/fdb963a1b46e11b5ba208c812ef7528ce7b549dc

PR #230, opened 2026-07-06
fix(nodeset): report PVC creation events with the create outcome

faganihajizada:
## Summary

In `createPersistentVolumeClaims`, the PVC create error was declared in the if statement header (`if err := r.Create(ctx, &claim); err != nil`), so the event condition below it kept evaluating the outer `r.Get` error, which in the `IsNotFound` case branch is always non-nil. As a result every successful PVC creation emitted a `Warning FailedCreate` event claiming the claim was not found, the `AlreadyExists` event suppression on the next line was dead code, and genuine create failures were reported with the stale NotFound text instead of the actual error. Reproduced on a kind cluster with the published `ghcr.io/slinkyproject/slurm-operator:1.2.0-rc1` image: creating a NodeSet with a `volumeClaimTemplate` produced the event `Warning FailedCreate create Claim: scratch-slurm-worker-pvc-0 for Pod slurm-worker-pvc-0 failed: PersistentVolumeClaim "scratch-slurm-worker-pvc-0" not found` while that same PVC had just been created successfully.

The fix restores that form: `err := r.Create(ctx, &claim)` at case block scope, so the error aggregation and the event condition both observe the create result. Successful creations now emit `Normal SuccessfulCreate`, a create that races with another creator and gets `AlreadyExists` emits no event (the error is still aggregated and returned, matching upstream), and real create failures are reported with the real error.

## Checklist

## Breaking Changes

N/A

## Testing Notes

The `createPersistentVolumeClaims` table test now asserts recorded events in every case: successful creation emits `Normal SuccessfulCreate`, a failing Get and a failing Create each emit `Warning FailedCreate` with the real error in the message, an `AlreadyExists` create is suppressed (new case, with the error still returned), and every case asserts no further events remain buffered. Three of these cases fail against the previous code, which emitted the stale NotFound warning in all of them. The full test suite passes.

Live on kind: with the published `1.2.0-rc1` operator image, a NodeSet with a `volumeClaimTemplate` produced the misleading `Warning FailedCreate ... not found` event for a successfully created PVC; after swapping only the operator image to this branch's build and recreating the same NodeSet, the identical scenario produced `Normal SuccessfulCreate create Claim: scratch-slurm-worker-pvc-0 Pod slurm-worker-pvc-0`.

### DIFF
--- a/internal/controller/nodeset/podcontrol/podcontrol.go
+++ b/internal/controller/nodeset/podcontrol/podcontrol.go
@@ -296,7 +296,8 @@ func (r *realPodControl) createPersistentVolumeClaims(ctx context.Context, nodes
 		err := r.Get(ctx, pvcId, pvc)
 		switch {
 		case apierrors.IsNotFound(err):
-			if err := r.Create(ctx, &claim); err != nil {
+			err := r.Create(ctx, &claim)
+			if err != nil {
 				errs = append(errs, fmt.Errorf("failed to create PVC %s: %w", claim.Name, err))
 			}
 			if err == nil || !apierrors.IsAlreadyExists(err) {
--- a/internal/controller/nodeset/podcontrol/podcontrol_test.go
+++ b/internal/controller/nodeset/podcontrol/podcontrol_test.go
@@ -9,9 +9,11 @@ import (
 	"fmt"
 	"net/http"
 	"reflect"
+	"strings"
 	"testing"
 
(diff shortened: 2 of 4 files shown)


vivian-hafener:
Merged: https://github.com/SlinkyProject/slurm-operator/commit/c48713c6abc22794d87ffa5aa41ffd2df869a85e

PR #241, opened 2026-07-21
fix(token): keep refresh requeue positive so refresh never stops

### LINKED ISSUE #240: Token controller stops refreshing (RequeueAfter ≤ 0 kills self-requeue); JWTs silently expire
### Version

1.1.0

### Which installation method(s) does this occur on?

Helm / Kubernetes manifests

### Describe the bug.

**Describe the bug**

The Token controller's periodic refresh can silently stop. On the reconcile scheduled at the refresh instant,
`Sync()` pushes `RequeueAfter = refreshTime.Sub(now)` where `refreshTime = exp − Lifetime/5` — so it is `≤ 0` at
that moment. `durationStore` uses `Less` (keeps the minimum, no floor), so `Pop` returns `≤ 0`, and controller-runtime
treats `RequeueAfter ≤ 0` as "do not requeue." The reconcile still mints a token and writes the Secret, so the loop
continues **only** via the `Owns(&Secret{})` watch follow-up. When that watch event is coalesced/dropped, nothing
re-enqueues the Token: it is never refreshed again, the JWT expires, and consumers get `Protocol authentication error`.
It self-recovers only on the ~10h cache resync or an operator restart, so tokens flap.

Introduced by **`f9d0a98` ("feat(token): ensure Token is refreshed timely", first shipped in v1.1.0)**, which added
`durationStore.Push(key, refreshTime.Sub(now))` and switched the store from `Greater` to `Less`. **v1.0.1 is unaffected.**

**Steps/Code to reproduce bug**

1. Create a `Token` with `refresh: true` (default 15m lifetime).
2. In the token-controller logs, each cycle emits **two** `Finished syncing Token` lines — one with no `result`
   (the mint, `RequeueAfter ≤ 0`) and one with `result.RequeueAfter ≈ +12m` (the `Owns(Secret)` follow-up).
3. Eventually a cycle emits only the no-result reconcile (the follow-up watch event is coalesced). After that there are
   no further Token reconciles until the ~10h resync; the Secret's JWT expires and stops being refreshed.

**Expected behavior**

A `Token` with `refresh: true` never expires. The controller should arm its own next refresh on every reconcile —
compute the requeue from the **newly-minted** token's `exp`, and never return `RequeueAfter ≤ 0` (floor to a positive
minimum, or set `Requeue: true`) — removing the dependency on the `Owns(Secret)` watch.

**Environment details (please complete the following information):**

- Environment location: Cloud (Kubernetes)
- Install / deployment method: Helm; slurm-operator **v1.1.0** (also present on current `main` and `v1.1.1`)

**Additional context**

Relevant code: `internal/controller/token/token_sync.go` (the `refreshTime.Sub(now)` push),
`internal/controller/token/token_controller.go` (`res = reconcile.Result{RequeueAfter: durationStore.Pop(req.String())}`,
no floor), and `internal/utils/durationstore` (`Less`, stores `≤0` verbatim, no clamp). The fix is small and local to
the token controller.

### Minimum reproducible example

```shell

```

### Relevant log output

```shell

```

### Additional environment details

```shell

```

### Other/Misc.

_No response_

### Code of Conduct

- [x] I agree to follow Slurm Operator's Code of Conduct
- [x] I have searched the [open bugs](https://github.com/SlinkyProject/slurm-operator/issues?q=is%3Aopen+is%3Aissue+label%3Abug) and have found no duplicates for this bug report

somaz94:
## Summary

The Token controller can silently stop refreshing. On the reconcile scheduled at the refresh instant, Sync() pushes RequeueAfter = refreshTime.Sub(now) where refreshTime = exp - Lifetime/5, so it is <= 0 at that moment. durationStore uses Less (keeps the minimum, no floor), so Pop returns <= 0 and controller-runtime treats that as "do not requeue". The loop then continues only via the Owns(&Secret{}) watch follow-up; when that event is coalesced or dropped, the Token is never re-enqueued, the JWT expires, and consumers get a Protocol authentication error until the ~10h cache resync.

This floors the refresh requeue to a positive minimum and arms the next refresh from the freshly-minted token instead of depending on the Secret watch.

closes #240

## Checklist

- [x] I have read the
  and the

## Breaking Changes

None.

## Testing Notes

- envtest suite for internal/controller/token/... passes with no regression.
- Added TestRefreshRequeueAfter covering the floor invariant (negative, zero, and below-floor inputs never yield <= 0).
- golangci-lint run ./internal/controller/token/ reports 0 issues.

## Additional Context

Introduced by f9d0a98 ("feat(token): ensure Token is refreshed timely", first shipped in v1.1.0), which added the durationStore.Push(refreshTime.Sub(now)) and switched the store from Greater to Less. v1.0.1 is unaffected.

### DIFF
--- a/internal/controller/token/token_requeue_test.go
+++ b/internal/controller/token/token_requeue_test.go
@@ -1,0 +1,54 @@
+// SPDX-FileCopyrightText: Copyright (C) SchedMD LLC.
+// SPDX-License-Identifier: Apache-2.0
+
+package token
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestRefreshRequeueAfter(t *testing.T) {
+	tests := []struct {
+		name string
+		d    time.Duration
+		want time.Duration
+	}{
+		{
+			name: "negative is floored to the positive minimum",
+			d:    -1 * time.Minute,
+			want: tokenRefreshRequeueFloor,
+		},
+		{
+			name: "zero is floored to the positive minimum",
+			d:    0,
+			want: tokenRefreshRequeueFloor,
+		},
+		{
+			name: "positive below the floor is raised to the floor",
+			d:    500 * time.Millisecond,
+			want: tokenRefreshRequeueFloor,
+		},
+		{
+			name: "exactly the floor is kept",
+			d:    tokenRefreshRequeueFloor,
+			want: tokenRefreshRequeueFloor,
+		},
+		{
+			name: "positive above the floor is kept",
+			d:    12 * time.Minute,
+			want: 12 * time.Minute,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got := refreshRequeueAfter(tt.d)
+			require.Equal(t, tt.want, got)
+			// The core invariant of #240: the refresh requeue is never <= 0,
+			// which controller-runtime would treat as "do not requeue".
+			require.Greater(t, got, time.Duration(0))
+		})
+	}
+}
--- a/internal/controller/token/token_sync.go
+++ b/internal/controller/token/token_sync.go
@@ -24,6 +24,21 @@ import (
 	jwt "github.com/golang-jwt/jwt/v5"
 )
 
+// tokenRefreshRequeueFloor bounds the token refresh requeue to a positive
+// value. controller-runtime treats a Result.RequeueAfter <= 0 as "do not
+// requeue", so an unbounded refreshTime.Sub(now) that reaches zero silently
+// stops the periodic refresh and lets the JWT expire (#240).
+const tokenRefreshRequeueFloor = 1 * time.Second
+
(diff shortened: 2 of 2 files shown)


afuller-crusoe:
This fix looks appropriate, should fix the silent enqueue drop while keeping the spirit for [f9d0a98](https://github.com/SlinkyProject/slurm-operator/commit/f9d0a98003da37400e4501584219d7a2d6e12220) timely token refresh

somaz94:
Thanks for the review @afuller-crusoe! Flipped it out of draft — ready for a formal review/merge whenever you get a chance 🙏

somaz94:
Closing this one out — I see development happens on the internal GitLab and changes get applied there, so this PR isn't the right vehicle. Happy for the fix to be picked up on that side if it's useful. Thanks for taking a look.

SkylerMalinowski:
Yes, the main repository is in a private GitLab, which is mirrored to public GitHub where the community can get involved. GitHub PRs will be considered for inclusion. If nothing else, they can help the discussion of an fix or feature.

PR #250, opened 2026-08-31
fix(nodeset): tolerate NotFound on cordon/uncordon of vanished pods (fixes rolling-update wedge)

### LINKED ISSUE #249: [Bug]: NodeSet rolling update wedges: node left drained ("Pod is terminating") after pod-patch race, replacement pod never created (v1.2.1)
## Summary

On a NodeSet revision change (pod template update), the rolling update
path drains the Slurm node, deletes the old pod — and then the NodeSet
permanently fails to create the replacement pod. The Slurm node record
stays `DOWN+DRAIN+NOT_RESPONDING` with reason
`slurm-operator: Pod is terminating`, every subsequent reconcile
completes in ~1s without action, and the node is lost from the
partition until an operator manually runs
`scontrol update nodename=<node> state=undrain` — after which the
replacement pod is created within one reconcile (~30s).

Reproduced on **3 out of 3 rolling updates** in one day (different
nodes, different template changes: image tag bump, extra volume).
A plain `kubectl delete pod` of a NodeSet pod (no revision change)
does **not** wedge — recovery is clean.

## Environment

- slurm-operator **1.2.1** (`ghcr.io/slinkyproject/slurm-operator:1.2.1`, chart `slurm-1.2.1`)
- Slurm 26.05, dynamic node registration
- Kubernetes: k0s v1.35.6 (K8s 1.35)
- NodeSet: `ScalingMode: Daemonset`, one pod per node (1–3 nodes),
  `UpdateStrategy: RollingUpdate` (defaults)

## Steps to reproduce

1. NodeSet with daemonset scaling mode, ≥1 running pod, Slurm node
   registered and idle.
2. Change the pod template (e.g. bump the slurmd image tag) →
   `helm upgrade` / apply.
3. Watch the NodeSet events and `sinfo`.

## Observed

NodeSet events (single occurrence, representative):

```
Normal   RollingUpdate     nodeset-controller  Rolling update: replacing 1 old pod(s) with updated revision
Normal   RollingUpdate     nodeset-controller  Rolling update: replacing 1 old pod(s) with updated revision
Normal   SuccessfulDelete  nodeset-controller  Deleted pod: slurm-worker-cpu-hqfcd
Normal   ScalingUp         nodeset-controller  Creating 1 daemon Pod(s)
Warning  SyncFailed        nodeset-controller  Failed "NodeSetPods" step: failed to patch slurm/slurm-worker-cpu-hqfcd: pods "slurm-worker-cpu-hqfcd" not found
```

Slurm side (stays like this indefinitely):

```
$ sinfo -N -o "%N %t %E"
wrk-b7 drain* slurm-operator: Pod is terminating
$ scontrol show node wrk-b7 | grep State
   State=DOWN+DRAIN+DYNAMIC_NORM+NOT_RESPONDING
```

Operator log afterwards: `Started syncing NodeSet` / `Finished syncing
NodeSet` every 30s, duration ~1s, no pod creation, no further events.
Observed wedged for 10–15+ minutes (until manual intervention), so it
is not the ControllerExpectations 5-minute expiry self-healing.

## Expected

The replacement pod is created; transient NotFound on a pod that the
controller itself just deleted should not abort the sync step chain,
and the drain placed by the operator should not outlive the pod it was
placed for.

## Workaround

`scontrol update nodename=<node> state=undrain` — replacement pod is
created within ~30s. We automated this in our external watchdog: if a
Slurm node is drained with reason `Pod is terminating` **and** no
slurmd pod exists on that node, undrain it.

## Analysis (v1.2.1 sources)

The failing patch happens in the `NodeSetPods` step
(`internal/controller/nodeset/nodeset_sync.go:398`) on a reconcile that
still sees the just-deleted pod through the stale informer cache: the
daemonset branch of `syncNodeSetPods` decides both "node needs a daemon
pod" (hence the `ScalingUp` event) and keeps processing the cached,
already-deleted pod; a pod patch (uncordon/label path —
`syncPodUncordon`/`makePodCordon`/`UpdateNodeSetPod` all patch the pod
object) then fails with NotFound via
`internal/utils/objectutils/patch.go`, and the step returns an error —
seemingly before the creation is committed.

Two things combine into the deadlock:

1. NotFound on patching a pod the controller itself deleted is treated
   as a step failure (unlike `DeleteNodeSetPod`, which already
   tolerates `IsNotFound`).
2. After the failed step, subsequent reconciles never re-attempt the
   creation, while `syncSlurmNodeRecords`/`syncSlurmNodes` keep the
   drained node record (reason `Pod is terminating`) in place; clearing
   the drain externally immediately unblocks pod creation, which
   suggests the drained/NOT_RESPONDING Slurm record participates in the
   "should this node get a pod" decision on the daemonset path.

Happy to provide full operator logs at higher verbosity, and happy to
send a PR if you can confirm the intended semantics (tolerate NotFound
on pod patches for self-deleted pods, and/or lift the operator-owned
drain when the pod it referred to is gone).

AndreyZa:
## Summary

Fixes #249 — NodeSet rolling update wedges: the node is left drained
(`slurm-operator: Pod is terminating`) and the replacement pod is never
created until a manual `scontrol undrain`.

## Root cause

On the reconcile right after the rolling update deletes the old pod, the
stale informer cache can still list that pod. In `doPodScale`:

1. creation expectations are raised (`RaiseExpectations`) **before** the
   uncordon batch;
2. `syncPodUncordon` → `makePodUncordon` patches the cached, already
   deleted pod and fails with `NotFound`
   (`failed to patch <ns>/<pod>: pods "<pod>" not found`), aborting the
   step **before any create is attempted**;
3. the raised creation expectations are never released (the existing
   `skippedPods` compensation only runs when the create batch itself
   runs), so subsequent reconciles wait for creations that were never
   issued — the NodeSet wedges, and the operator-owned drain outlives
   the pod it was placed for.

Reproduced on 3 out of 3 rolling updates (daemonset scaling mode, Slurm
26.05, operator 1.2.1); details, events and logs in #249.

## Changes

- `makePodCordon` / `makePodUncordon`: treat `NotFound` as a no-op — the
  pod is gone, there is nothing to (un)cordon. This mirrors the existing
  handling in the deadline and topology pod patches in the same file.
- `doPodScale`: when the uncordon batch fails, release the creation
  expectations raised earlier in the same sync, so an aborted step can
  no longer block future creations.
- Tests: the two `NotFound` subtests of `makePodCordon`/`makePodUncordon`
  now expect success.

## Testing

- `go build ./...`, `gofmt`, `go vet` clean.
- `go test ./internal/controller/nodeset/...` green (envtest suite
  included, k8s 1.37 assets).
- The failure mode itself was exercised live on a test cluster (three
  wedged rollouts in one day before the fix; with NotFound tolerated the
  aborting patch disappears from the chain).

### DIFF
--- a/internal/controller/nodeset/nodeset_sync.go
+++ b/internal/controller/nodeset/nodeset_sync.go
@@ -1196,6 +1196,12 @@ func (r *NodeSetReconciler) doPodScale(
 		return r.syncPodUncordon(ctx, nodeset, pod)
 	}
 	if _, err := utils.SlowStartBatch(len(podsToKeep), utils.SlowStartInitialBatchSize, uncordonFn); err != nil {
+		// No creates will be attempted this sync; release the creation
+		// expectations raised above, otherwise subsequent reconciles wait
+		// for creations that were never issued (see issue #249).
+		for range numCreate {
+			r.expectations.CreationObserved(logger, key)
+		}
 		return err
 	}
 
@@ -1529,6 +1535,12 @@ func (r *NodeSetReconciler) makePodCordon(
 		return nil
 	}
 	if err := objectutils.PatchObject(r.Client, ctx, pod, mutateFn); err != nil {
+		// The pod may already be gone (e.g. deleted by this controller in a
+		// previous reconcile and still present in the informer cache).
+		// There is nothing left to cordon in that case.
+		if apierrors.IsNotFound(err) {
+			return nil
+		}
 		return err
 	}
 
@@ -1568,6 +1580,13 @@ func (r *NodeSetReconciler) makePodUncordon(ctx context.Context, pod *corev1.Pod
 		return nil
 	}
 	if err := objectutils.PatchObject(r.Client, ctx, pod, mutateFn); err != nil {
+		// The pod may already be gone (e.g. deleted by a rolling update in a
+		// previous reconcile and still present in the informer cache).
+		// Failing here would abort the whole NodeSetPods step before any
+		// replacement pods are created (see issue #249).
+		if apierrors.IsNotFound(err) {
+			return nil
+		}
 		return err
 	}
 
--- a/internal/controller/nodeset/nodeset_sync_test.go
+++ b/internal/controller/nodeset/nodeset_sync_test.go
@@ -2490,7 +2490,7 @@ func TestNodeSetReconciler_makePodCordon(t *testing.T) {
 				ctx: context.TODO(),
 				pod: pod1.DeepCopy(),
 			},
-			wantErr: true,
+			wantErr: false,
 		},
 		{
 			name: "cordoned",
@@ -2731,7 +2731,7 @@ func TestNodeSetReconciler_makePodUncordon(t *testing.T) {
 				ctx: context.TODO(),
 				pod: pod1.DeepCopy(),
 			},
-			wantErr: true,
+			wantErr: false,
 		},
 		{
 			name: "cordoned",


AndreyZa:
Closing: the failure this patch guarded against no longer reproduces on 1.2.2 — see #249 for the verification (two full rolling updates on a production cluster, both clean).

The change itself is still a small hardening (tolerating `NotFound` when cordoning or uncordoning a pod that has already vanished), so feel free to pick it up if you find it useful; otherwise no action needed. Happy to reopen with fresh evidence if the wedge returns.

PR #253, opened 2026-09-02
fix(builder): keep explicitly-set false/0 values when merging CR templates

### LINKED ISSUE #252: [Bug]: explicitly-set false/0 values in CR templates are dropped when building the workload (allowPrivilegeEscalation, privileged, runAsUser: 0, …)
## Summary

Values in a CR pod/container template that are equal to the zero value of
their Go type — `false`, `0`, `""` — are pruned before the template is
merged into the operator's base spec. The CR keeps the field, the built
Deployment/StatefulSet does not, and nothing is logged or surfaced in
status.

The most visible effect: `spec.login.securityContext.allowPrivilegeEscalation: false`
on a `LoginSet` never reaches the Deployment, so the login container runs
with `NoNewPrivs: 0` and a `LoginSet` cannot satisfy the `restricted` Pod
Security Standard no matter what the CR says. `privileged: false`,
`readOnlyRootFilesystem: false`, `runAsUser: 0` and
`terminationGracePeriodSeconds: 0` are lost the same way.

This also affects the chart's own defaults: `helm/slurm/values.yaml` ships
`loginsetDefaults.login.securityContext.privileged: false`, and that value
does not survive to the Deployment either.

The pruning happens in `internal/utils/structutils`, which every builder
uses (`BuildContainer`, `BuildPodTemplate`), so NodeSet, Controller,
Accounting and RestApi templates are affected identically.

## Environment

- slurm-operator **1.2.1** (`ghcr.io/slinkyproject/slurm-operator:1.2.1`,
  charts `slurm-operator-1.2.1`, `slurm-1.2.1`)
- Kubernetes v1.35.8 (k0s), single node
- Reproduced on a scratch cluster with a stock chart install; the same
  behaviour is present on `main` (`internal/utils/structutils/kube.go`
  is unchanged between `v1.2.1` and `main` as of 2026-09-02).

## Steps to reproduce

Install the charts and add one `LoginSet` through values:

```yaml
loginsets:
  probe:
    enabled: true
    login:
      securityContext:
        privileged: false
        allowPrivilegeEscalation: false
        readOnlyRootFilesystem: false
        capabilities:
          drop: [ALL]
          add: [CHOWN, SETUID, SETGID, SYS_CHROOT]
    podSpec:
      terminationGracePeriodSeconds: 0
      securityContext:
        runAsUser: 0
      nodeSelector:
        kubernetes.io/os: linux
        ss-probe/marker: "yes"
```

Then compare the CR with what the operator built:

```console
$ kubectl -n slurm get loginset slurm-login-probe \
    -o jsonpath='{.spec.login.securityContext}'
{"allowPrivilegeEscalation":false,"capabilities":{"add":["CHOWN","SETUID","SETGID","SYS_CHROOT"],"drop":["ALL"]},"privileged":false,"readOnlyRootFilesystem":false}

$ kubectl -n slurm get loginset slurm-login-probe \
    -o jsonpath='{.spec.template.spec.terminationGracePeriodSeconds} {.spec.template.spec.securityContext}'
0 {"runAsUser":0}
```

## Observed (operator 1.2.1)

```console
$ kubectl -n slurm get deploy slurm-login-probe \
    -o jsonpath='{.spec.template.spec.containers[?(@.name=="login")].securityContext}'
{"capabilities":{"add":["CHOWN","SETUID","SETGID","SYS_CHROOT"],"drop":["ALL"]}}

$ kubectl -n slurm get deploy slurm-login-probe \
    -o jsonpath='{.spec.template.spec.terminationGracePeriodSeconds} {.spec.template.spec.securityContext}'
30 {}
```

`allowPrivilegeEscalation`, `privileged`, `readOnlyRootFilesystem`,
`runAsUser: 0` are gone; `terminationGracePeriodSeconds: 0` was replaced
by the API default of 30. Inside the pod: `NoNewPrivs: 0`.

Non-zero values in the very same objects (capabilities, images,
resources, `nodeSelector`) are propagated correctly — only zero values
vanish.

## Expected

A field explicitly present in the CR template reaches the built workload
verbatim. In the Kubernetes API `false` and `0` are choices, not
"unset": `allowPrivilegeEscalation: false` is required by the
`restricted` Pod Security Standard, `privileged: false` is the only way
to override a `true` coming from a base spec (relevant to #238), and
`runAsUser: 0` is an explicit request for root.

## Root cause (v1.2.1 and main)

`internal/builder/common/container.go` and
`internal/builder/common/pod_template.go` both merge the user template
through `structutils.StrategicMergePatch`, which cleans the patch first:

```go
// internal/utils/structutils/kube.go
patchBytes, err := cleanAndMarshal(patch)
...
func removeEmpty(m map[string]any) {
    for k, v := range m {
        if v == nil || reflectutils.IsEmpty(v) {   // IsEmpty: DeepEqual(v, zero(typeof v))
            delete(m, k)
        } else if subMap, ok := v.(map[string]any); ok {
            removeEmpty(subMap)
            if len(subMap) == 0 {
                delete(m, k)
            }
        }
    }
}
```

After `json.Unmarshal` into `map[string]any`, a JSON `false` is a Go
`bool` whose zero value is `false`, and a JSON `0` is a `float64` zero,
so `IsEmpty` reports both as empty and `removeEmpty` deletes the key.

The intent — not letting unset fields clobber the base — is sound, but
the fields at risk are `*bool`/`*int64` with `omitempty` in the
Kubernetes API types: they are marshalled **only** when the user set
them. By the time the patch is a generic map, "unset" and "explicitly
false" are indistinguishable, and pruning discards configuration rather
than noise.

## Fix

Prune only values that genuinely carry no information — nulls, empty
strings, empty objects, empty lists — and keep booleans and numbers.
PR: #253 (with a regression test covering container- and pod-level zero
values; the test fails on `main` and passes with the change).

Verified on the scratch cluster above by swapping only the operator
image (chart, webhook and CRs untouched): all five fields appear in the
Deployment and the login pod reports `NoNewPrivs: 1`.

## Workaround for users on 1.2.1

A `MutatingAdmissionPolicy` matched on the login pods that re-injects
`securityContext.allowPrivilegeEscalation: false` on CREATE. It works,
but it is a second source of truth for a field the CR already declares
and it is invisible to anyone reading the `LoginSet`.

AndreyZa:
## Summary

`removeEmpty()` pruned every JSON value equal to the zero value of its Go type, so an explicitly-set `false` or `0` in a CR pod/container template was dropped from the strategic merge patch and never reached the built workload.

The fields at risk are `*bool`/`*int64` with `omitempty` in the Kubernetes API types: they are marshalled **only** when the user set them, so pruning them discarded configuration rather than noise. Nulls, empty strings and empty objects are still pruned, and lists are left exactly as `main` leaves them — an explicitly empty array is preserved — so nothing beyond booleans and numbers changes.

Concretely, `spec.login.securityContext.allowPrivilegeEscalation: false` on a `LoginSet` is stored in the CR but absent from the Deployment, so the login container runs with `NoNewPrivs: 0` and a `LoginSet` cannot satisfy the `restricted` Pod Security Standard. The chart's own default (`loginsetDefaults.login.securityContext.privileged: false`) is lost the same way, as are `readOnlyRootFilesystem: false`, `runAsUser: 0` and `terminationGracePeriodSeconds: 0`.

One clarification on the scope, since the helper is shared by `BuildContainer` and `BuildPodTemplate`: the loss happens at the **root of the merge** — the container itself for `BuildContainer`, and pod-level fields for `BuildPodTemplate`. `removeEmpty` never descended into lists, so a container sitting inside `spec.containers[]` already came through intact on `main`.

Closes #252.

## Checklist

- [x] I have read the
  and the
- [x] Documentation is updated if user-visible behavior changes. (No documentation change: the fix makes the builders honour values the CRD already documents.)

## Breaking Changes

No API or schema change, but the fix is **retroactive**, and that is worth stating plainly: a `false` or `0` already sitting in a CR or in chart values is inert today and becomes effective the moment the operator image is replaced. No CR is edited, no chart value changes, nothing shows up in a diff — the same configuration simply starts being honoured.

Mostly that is the point, and it hardens: `allowPrivilegeEscalation: false`, `privileged: false` and `hostUsers: false` start reaching the pod. Two directions deserve a look before upgrading:

* the base specs of Controller, Accounting and RestApi set `runAsNonRoot: true` with a non-root uid; an explicit `runAsNonRoot: false` or `runAsUser: 0` in a CR now overrides that, where before the operator silently won;
* `terminationGracePeriodSeconds: 0` now reaches the pod, which removes the window the slurmd `preStop` hook uses to drain the node.

A grep for zero-valued `securityContext` fields and for `terminationGracePeriodSeconds` across CRs and values is enough to see whether a given deployment is affected. A deployment that relied on an explicit `false` being silently ignored would change, but that would be the bug itself.

## Testing Notes

* `internal/utils/structutils/kube_test.go` gains a subtest that merges at both roots the builders use. A pod template carrying `automountServiceAccountToken: false`, `terminationGracePeriodSeconds: 0` and `securityContext.runAsUser: 0`, and a container — the root `BuildContainer` merges at — carrying `privileged: false`, `allowPrivilegeEscalation: false` and `readOnlyRootFilesystem: false` over a base that sets them `true`. Both halves fail on `main` and pass with this change. `make test` is green (50 packages) and `golangci-lint run ./internal/...` reports nothing from these files.
* List handling is deliberately untouched: `removeEmpty` has no list branch at all, so an explicitly empty array survives here exactly as it does on `main`.
* A known limit this does not change: non-pointer `bool` fields such as `hostNetwork`, `hostPID` and `hostIPC`, and non-pointer empty strings, are dropped by `omitempty` during `json.Marshal`, before `removeEmpty` ever sees them. No base spec sets those to `true` today, so there is nothing to override, but the class is not fully gone.
* Verified on a cluster (Kubernetes v1.35.8, charts `slurm-operator-1.2.1` / `slurm-1.2.1`, one `LoginSet` declaring all five fields; only the operator image swapped):

  Before (`ghcr.io/slinkyproject/slurm-operator:1.2.1`):

  ```
  container.securityContext: {"capabilities":{"add":[...],"drop":["ALL"]}}
  pod.terminationGracePeriodSeconds: 30
  pod.securityContext: {}
  ```

  After (this branch):

  ```
  container.securityContext: {"allowPrivilegeEscalation":false,"capabilities":{"add":[...],"drop":["ALL"]},"privileged":false,"readOnlyRootFilesystem":false}
  pod.terminationGracePeriodSeconds: 0
  pod.securityContext: {"runAsUser":0}
  ```

  and in the login pod: `NoNewPrivs: 1`.

## Additional Context

Found while trying to run a `LoginSet` under the `restricted` Pod Security Standard.

### DIFF
--- a/internal/utils/structutils/kube.go
+++ b/internal/utils/structutils/kube.go
@@ -7,8 +7,6 @@ import (
 	"encoding/json"
 
 	"k8s.io/apimachinery/pkg/util/strategicpatch"
-
-	"github.com/SlinkyProject/slurm-operator/internal/utils/reflectutils"
 )
 
 // StrategicMergePatch merges two objects via kubernetes StrategicMergePatch
@@ -65,10 +63,11 @@ func cleanAndMarshal(obj any) ([]byte, error) {
 	return out, nil
 }
 
-// removeEmpty will recursively walk a map, deleting fields if its value is empty.
+// removeEmpty will recursively walk a map, deleting fields that carry no
+// information: nulls, empty strings and empty objects.
 func removeEmpty(m map[string]any) {
 	for k, v := range m {
-		if v == nil || reflectutils.IsEmpty(v) {
+		if v == nil || v == "" {
 			delete(m, k)
 		} else if subMap, ok := v.(map[string]any); ok {
 			removeEmpty(subMap)
--- a/internal/utils/structutils/kube_test.go
+++ b/internal/utils/structutils/kube_test.go
@@ -9,12 +9,78 @@ import (
 	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
 
 	"github.com/SlinkyProject/slurm-operator/internal/utils/structutils"
 )
 
 func Test_strategicMergePatch(t *testing.T) {
 	test_strategicMergePatch_pod(t)
+	test_strategicMergePatch_explicitZeroValues(t)
+}
+
+// Explicitly-set zero values are meaningful in the Kubernetes API and must
+// survive the merge: `allowPrivilegeEscalation: false` is required by the
+// restricted Pod Security Standard, `privileged: false` is the only way to
+// override a `true` in the base spec, and `runAsUser: 0` is an explicit
+// request for root.
+func test_strategicMergePatch_explicitZeroValues(t *testing.T) {
+	// The pod template path, as used by BuildPodTemplate.
+	base := &corev1.Pod{
+		Spec: corev1.PodSpec{
+			Containers: []corev1.Container{
+				{
+					Name:  "app",
+					Image: "app",
+				},
+			},
+		},
+	}
+	patch := &corev1.Pod{
+		Spec: corev1.PodSpec{
+			AutomountServiceAccountToken:  ptr.To(false),
+			TerminationGracePeriodSeconds: ptr.To(int64(0)),
+			SecurityContext: &corev1.PodSecurityContext{
+				RunAsUser: ptr.To(int64(0)),
+			},
+		},
+	}
+
+	got := structutils.StrategicMergePatch(base, patch)
+
+	require.Equal(t, ptr.To(false), got.Spec.AutomountServiceAccountToken)
+	require.Equal(t, ptr.To(int64(0)), got.Spec.TerminationGracePeriodSeconds)
+	require.NotNil(t, got.Spec.SecurityContext)
+	require.Equal(t, ptr.To(int64(0)), got.Spec.SecurityContext.RunAsUser)
+	// Fields only present in the base are untouched by the patch.
+	require.Len(t, got.Spec.Containers, 1)
+	require.Equal(t, "app", got.Spec.Containers[0].Image)
+
+	// The container path, as used by BuildContainer, where the container is
+	// the root of the merge.
+	baseContainer := &corev1.Container{
+		Name:  "login",
+		Image: "login",
+		SecurityContext: &corev1.SecurityContext{
+			Privileged:               ptr.To(true),
+			AllowPrivilegeEscalation: ptr.To(true),
+		},
+	}
+	patchContainer := &corev1.Container{
+		SecurityContext: &corev1.SecurityContext{
+			Privileged:               ptr.To(false),
+			AllowPrivilegeEscalation: ptr.To(false),
+			ReadOnlyRootFilesystem:   ptr.To(false),
+		},
+	}
+
+	gotContainer := structutils.StrategicMergePatch(baseContainer, patchContainer)
+
+	require.NotNil(t, gotContainer.SecurityContext)
+	require.Equal(t, ptr.To(false), gotContainer.SecurityContext.AllowPrivilegeEscalation)
+	require.Equal(t, ptr.To(false), gotContainer.SecurityContext.Privileged)
+	require.Equal(t, ptr.To(false), gotContainer.SecurityContext.ReadOnlyRootFilesystem)
+	require.Equal(t, "login", gotContainer.Image)
 }
 
 func test_strategicMergePatch_pod(t *testing.T) {


vivian-hafener:
Please update this PR to use the provided Pull Request Template: https://github.com/SlinkyProject/slurm-operator/blob/main/.github/pull_request_template.md

AndreyZa:
Hi @vivian-hafener — thanks for the pointer, and sorry for the silence here: I rewrote the description to follow the template the same day (Sep 8), but never said so in the thread.

The body now has the template's Summary / Checklist / Breaking Changes / Testing Notes / Additional Context sections. No code changes since it was opened — the branch is ready for review whenever you have a moment.

internal/utils/structutils/kube.go
```diff
@@ -65,14 +63,33 @@ func cleanAndMarshal(obj any) ([]byte, error) {
 	return out, nil
 }
 
-// removeEmpty will recursively walk a map, deleting fields if its value is empty.
+// removeEmpty will recursively walk a map, deleting fields that carry no
+// information: nulls, empty strings, empty objects and empty lists.
+//
+// Booleans and numbers are deliberately kept, even when they are the zero
```
vivian-hafener, line 69: Comment should be more concise - just lines 66-67 is sufficient.

internal/utils/structutils/kube.go
vivian-hafener, line 97: I would prefer a simpler implementation, something along the lines of this should work:
```go
func removeEmpty(m map[string]any) {
      for k, v := range m {
          if v == nil || v == "" {
              delete(m, k)
          } else if subMap, ok := v.(map[string]any); ok {
              removeEmpty(subMap)
              if len(subMap) == 0 {
                  delete(m, k)
              }
          }
      }
  }
```

internal/utils/structutils/kube.go
```diff
@@ -65,14 +63,33 @@ func cleanAndMarshal(obj any) ([]byte, error) {
...
+		case nil:
 			delete(m, k)
-		} else if subMap, ok := v.(map[string]any); ok {
-			removeEmpty(subMap)
-			if len(subMap) == 0 {
+		case string:
+			if value == "" {
+				delete(m, k)
+			}
+		case []any:
```
vivian-hafener, line 86: `case []any` is an unrelated behavioral change - `main` preserves explicit empty arrays, but this PR prunes them.

internal/utils/structutils/kube_test.go
```diff
@@ -9,12 +9,67 @@ import (
...
 	test_strategicMergePatch_pod(t)
+	test_strategicMergePatch_explicitZeroValues(t)
+}
+
+// Explicitly-set zero values (false, 0, and a zero quantity) are meaningful in
+// the Kubernetes API and must survive the merge: `allowPrivilegeEscalation:
+// false` is required by the restricted Pod Security Standard, `privileged:
+// false` is the only way to override a `true` in the base spec, and
+// `runAsUser: 0` is an explicit request for root.
+func test_strategicMergePatch_explicitZeroValues(t *testing.T) {
```
vivian-hafener: I believe that this regression test does not actually exercise the implemented behavior: it passes on `main` without your fix. This is likely because `removeEmpty` does not recurse into `[]any`, and so is not traversing down into `Pod.Spec.Containers`. Instead of using a `Pod` for the patch, try using a `Container`.

vivian-hafener (changes requested):

internal/utils/structutils/kube.go
AndreyZa: Done — the comment is now just those two lines. The rationale for keeping booleans and numbers moved to the commit message, where it does not have to be re-read on every visit to the function.

internal/utils/structutils/kube.go
AndreyZa: Adopted verbatim — `removeEmpty` is now your version, down to the variable names.

I checked that it is behaviourally identical to `main` apart from the booleans and numbers this PR is about, by running both versions over the same inputs: every JSON shape reachable after `json.Unmarshal` into `map[string]any` (null, bool, number, string, list, object, and those nested in each other). The only keys that differ are the zero-valued booleans and numbers. `v == ""` on an `any` is also safe for the uncomparable dynamic types — `[]any` and `map[string]any` compare false rather than panicking, since the dynamic types differ.

internal/utils/structutils/kube.go
AndreyZa: You were right, and I had it backwards in the description. Confirmed by running `main`'s `removeEmpty` over a map containing `[]`: `IsEmpty` compares the value against the zero value of `[]interface{}`, which is a nil slice, and `reflect.DeepEqual([]any{}, []any(nil))` is false — so an explicitly empty array survived on `main`, and my `case []any` was pruning it.

That branch is gone; there is no list handling left in the function at all. I also corrected the PR description, which still claimed empty lists were pruned.

AndreyZa:
Force-pushed a revised commit (`cdc1e83` → `7275462`) with all three points addressed; replies in the threads.

Two things beyond the review, both found while re-checking the patch:

**The test was half-decorative.** Its container assertions lived inside `Pod.Spec.Containers[]`, and `removeEmpty` never recursed into lists — not on `main` either — so those assertions passed with and without the fix. They are now a second merge at the root `BuildContainer` actually uses, a bare `corev1.Container`, which does fail on `main`. That is also the case the description leads with (`allowPrivilegeEscalation: false` on a login container), and it covers `readOnlyRootFilesystem: false` as well. The pod-template half is unchanged and still fails on `main`.

**The `Changelog` trailer would have failed CI.** It was a single 149-character line, over `body-max-line-length`; the repo's commitlint hook rejects that. It is now folded across continuation lines — `git interpret-trailers --parse` still reads it as one trailer, and `commitlint --strict` exits 0.

`make test` is green (50 packages). `golangci-lint run ./internal/...` is clean except for a pre-existing `misspell` in `internal/controller/token/token_controller_test.go`, which `main` has already fixed — this branch is based on an older commit and does not touch that file. The two files this PR does touch have not moved on `main` since the branch point, so it still applies cleanly; happy to rebase if you would rather have it on current `main`.

vivian-hafener:
Thanks for the fixes! Please rebase against `main`, then I'll take another look.

PR #254, opened 2026-09-02
feat(chart): default sshd to sending client keepalives

giuliocalzo:
Long-lived SSH sessions to login nodes freeze after a few minutes of inactivity.

The operator mounts its generated `sshd_config` over `/etc/ssh/sshd_config`, so no `ClientAlive*` setting is ever in effect and sshd never probes its clients. The login service defaults to `LoadBalancer`, and the provider's idle timeout (350s on AWS NLB, 4 minutes on Azure) then drops the connection silently. `TCPKeepAlive` does not help, since Linux only starts probing after two hours.

This defaults `loginsetDefaults.extraSshdConfig` and `nodesetDefaults.ssh.extraSshdConfig` to `ClientAliveInterval 60` / `ClientAliveCountMax 3`. Values-only: no Go, API, or CRD change, and a user-supplied `extraSshdConfig` still replaces it. `ClientAlive*` does not collide with any keyword the operator already emits.

To be explicit about the trade-off: this is a stability improvement, not a fix for something broken, and it is arguably unnecessary if we want the chart to stay as neutral as possible about its environment. The input comes from a real case with long-lived SSH connections, which is why I am proposing it rather than assuming it. Happy to drop it if chart neutrality wins.

`helm lint` and `helm unittest` pass (180 tests, including four new cases covering the default on both CRs and a user override); `README.md` and the LoginSet snapshot are regenerated.


AndreyZa:
Data point in favour of this default, from a small production cluster (Slinky 1.2.1, Slurm 26.05, k0s).

Our login service is `NodePort` behind NAT rather than a cloud `LoadBalancer`, and we hit exactly the same symptom: users' interactive sessions froze after a few minutes of inactivity, because the NAT idle timeout dropped the connection and sshd, with the operator-generated `sshd_config`, never sent a keepalive. So the problem is not specific to cloud load balancers; any stateful middlebox between client and login pod produces it, and the chart default silently being "no keepalives" is what surprises people.

We fixed it the same way you propose, via `extraSshdConfig`, currently with `ClientAliveInterval 60` and `ClientAliveCountMax 15`. The higher `CountMax` is deliberate for the login set: laptops sleep and VPNs flap, and 15 minutes of tolerance keeps those sessions alive, whereas 3 minutes would reap them. For the nodeset SSH the opposite trade-off (reaping dead sessions quickly so they do not pin the adopted job) makes 3 sensible, so you may want different defaults for the two CRs. Either way, having a non-null default is a clear improvement over the current behaviour, and it is easy to override.

PR #256, opened 2026-09-04
docs: document sshd configuration and client keepalives

giuliocalzo:
Documents the generated `sshd_config` and how to extend it, instead of changing any chart defaults. Supersedes #254, which proposed baking keepalives into `values.yaml`; the review preference upstream was to keep all `extra*` fields empty and document the configuration, which is what this does.

Adds `docs/usage/ssh-access.md`, covering:

- **How `sshd_config` is assembled.** The operator generates the whole file and mounts it over `/etc/ssh/sshd_config`, so the image's own defaults do not apply — worth stating, since it is invisible from the chart.
- **Client keepalives.** Nothing sets `ClientAliveInterval`, so it stays at OpenSSH's `0` and sshd never probes its clients. Long-lived sessions are then dropped by the idle timeout of whatever sits in front of them (350s AWS NLB, 600s GCP, 4 min Azure; `TCPKeepAlive` does not help, as Linux waits two hours). The reverse case is a vanished client leaving a session that, on a worker, still holds its adopted job. Documented with the reasoning for choosing an interval rather than just a snippet to paste.
- **Settings the operator already defines.** `sshd_config` takes the *first* value per keyword and `extraSshdConfig` is appended last, so keywords the operator already emits cannot be overridden there — the added line is silently ignored. The page lists them for login and worker pods and points at an `Include` drop-in for changing them. This is also why `ClientAlive*` is legitimately an `extraSshdConfig` setting: it does not collide with anything.

Docs only, no chart or code change. Formatted with the repo's `mdformat --wrap=80` pre-commit hook; the page is picked up automatically by the `usage/*` toctree glob.

### DIFF
--- a/docs/usage/ssh-access.md
+++ b/docs/usage/ssh-access.md
@@ -1,0 +1,73 @@
+# SSH Access
+
+## Table of Contents
+
+<!-- mdformat-toc start --slug=github --no-anchors --maxlevel=6 --minlevel=1 -->
+
+- [SSH Access](#ssh-access)
+  - [Table of Contents](#table-of-contents)
+  - [Generated sshd_config](#generated-sshd_config)
+  - [Client Keepalives](#client-keepalives)
+  - [Overriding Operator Settings](#overriding-operator-settings)
+
+<!-- mdformat-toc end -->
+
+## Generated sshd_config
+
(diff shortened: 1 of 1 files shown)


SkylerMalinowski:
Merged: 334c26f9

PR #258, opened 2026-09-04
fix(controller): keep labeling the active controller when Slurm is unreachable

### LINKED ISSUE #257: [Bug]: Controller Service loses all endpoints after 1.2.2 upgrade: the active-controller label is never set when Slurm is unreachable
## Summary

Upgrading to 1.2.2 leaves the cluster headless: the `Controller` Service ends up with **zero endpoints** and nothing recovers it. Every component that talks to `slurmctld` through that Service — `slurmd`, login pods, `slurmrestd`, and the operator itself — fails until an operator labels a controller pod by hand.

Chart 1.2.2 added `controller.slinky.slurm.net/active: "true"` to the Service selector (`internal/builder/controllerbuilder/controller_service.go`). The only code that sets that label is `syncHAStatus` (`internal/controller/controller/controller_sync_status.go`), and it sets it *after* asking Slurm which controller is active:

```go
pings, err := r.slurmControl.GetActiveHAController(ctx, controller)
if err != nil && !errors.Is(err, slurmcontrol.ErrNoSlurmClient) {
    return err
}
```

`GetActiveHAController` goes through the Slurm client → `slurmrestd` → the `slurm-controller` Service. If that Service has no endpoints the call fails with a transport error, not `ErrNoSlurmClient`, so `syncHAStatus` returns before labeling — and the Service stays without endpoints. The loop is closed:

> no label → no endpoints → Slurm unreachable → ping fails → return before labeling → no label

The pod template does not carry the label (`controller_app.go` builds it from `WithControllerLabels`), so a pod that is recreated for any reason starts without it.

It does not heal by itself. The Slurm client is created once and kept (`internal/controller/slurmclient/slurmclient_sync.go` updates it in place, never removes it), so `ErrNoSlurmClient` — the one error the code tolerates — stops being returned as soon as a client has been created. `Started syncing Controller` / `Finished syncing Controller` keep appearing in the log, and the reconciler reports no error above the default verbosity; the only visible symptom is downstream noise like `[Bad Gateway, Unable to contact slurm controller (connect failure)]` from the NodeSet reconciler.

## Environment

- operator **1.2.2**, charts `slurm-operator-1.2.2`, `slurm-1.2.2` (also reproduced on a scratch cluster)
- Slurm 26.05.4, single controller replica, `external: false`
- Kubernetes v1.35.8 (k0s), 3 control-plane VMs + 4 workers

## Steps to reproduce

Upgrade an existing 1.2.1 install to 1.2.2, or reproduce directly on a healthy 1.2.2 cluster:

```console
$ kubectl -n slurm get endpoints slurm-controller
NAME               ENDPOINTS            AGE
slurm-controller   10.244.0.253:6817    9d

$ kubectl -n slurm label pod slurm-controller-0 controller.slinky.slurm.net/active-
pod/slurm-controller-0 unlabeled
```

Then watch for as long as you like — three minutes shown here, the same happens over an hour:

```console
08:11:51 label=<none> endpoint=<none>
08:12:11 label=<none> endpoint=<none>
...
08:14:33 label=<none> endpoint=<none>
```

`sinfo` from inside the controller pod still works (slurmctld is healthy), while every client of the Service fails:

```console
$ kubectl -n slurm exec slurm-worker-0 -c slurmd -- sinfo
slurm_load_partitions: Unable to contact slurm controller (connect failure)
```

Recovery is a single command, and it is the only thing that works:

```console
$ kubectl -n slurm label pod slurm-controller-0 controller.slinky.slurm.net/active=true --overwrite
$ kubectl -n slurm get endpoints slurm-controller
NAME               ENDPOINTS           AGE
slurm-controller   10.244.0.253:6817   9d
```

## Impact

The chart upgrade itself triggers it: the controller pod is recreated with the new revision, comes up without the label, and the new Service selector no longer matches it. On our cluster this took out slurmd on all nodes, both login pods, `slurmrestd` and the operator's own Slurm access until we labeled the pod. Any later recreation of the controller pod (node drain, rollout, eviction) reproduces it.

## Suggested fix

Treat a failed ping like a missing client: fall through with no pings, label the primary, and let a later reconcile move the label once Slurm answers again. That keeps the Service reachable, which is the precondition for ever learning who is active.

AndreyZa:
Fixes #257.

`syncHAStatus` is the only place that sets `controller.slinky.slurm.net/active`, which chart 1.2.2 added to the `Controller` Service selector. It set the label only after `GetActiveHAController` succeeded, and that call goes through the very Service it is about to populate:

> no label → Service has no endpoints → Slurm unreachable → ping returns a transport error (not `ErrNoSlurmClient`) → `syncHAStatus` returns before labeling → no label

Once a Slurm client has been created it is kept for the lifetime of the operator (`slurmclient_sync.go` updates it in place), so `ErrNoSlurmClient` — the single error the code tolerated — stops being returned, and the cluster stays headless until someone labels a pod by hand. We hit this on a 1.2.1 → 1.2.2 chart upgrade: the controller pod was recreated without the label, and slurmd on every node, both login pods, slurmrestd and the operator's own Slurm access were down until we ran `kubectl label`.

This treats any ping failure the same way a missing client is already treated: fall through with no pings, label the primary, and let a later reconcile move the label once Slurm answers again. Labeling the primary is what the code already does for `ErrNoSlurmClient` and for the "nobody responded" case, so no new policy is introduced; the alternative — no endpoints at all — cannot recover on its own.

Testing:

- `go test ./internal/controller/controller/... -run TestControllerReconciler_syncHAStatus` — the existing case that asserted the error was propagated becomes a regression test asserting the primary gets labeled instead.
- Reproduced the deadlock on a 1.2.2 cluster (remove the label, watch the Service stay empty for minutes), then confirmed the patched behaviour restores the endpoint on the next reconcile.

### DIFF
--- a/internal/controller/controller/controller_sync_status.go
+++ b/internal/controller/controller/controller_sync_status.go
@@ -103,9 +103,14 @@ func (r *ControllerReconciler) syncHAStatus(
 		return nil
 	}
 
+	errs := []error{}
+
 	pings, err := r.slurmControl.GetActiveHAController(ctx, controller)
 	if err != nil && !errors.Is(err, slurmcontrol.ErrNoSlurmClient) {
-		return err
+		// Do not bail out when Slurm cannot be asked which controller is active.
+		log.FromContext(ctx).V(1).Info("Failed to determine the active controller, defaulting to the primary",
+			"Controller", klog.KObj(controller), "error", err)
+		errs = append(errs, fmt.Errorf("failed to determine the active controller: %w", err))
 	}
 
 	activePodName := controller.PodName(0)
@@ -126,7 +131,6 @@ func (r *ControllerReconciler) syncHAStatus(
 	}
 	sort.Sort(objectutils.PodsByName(podList.Items))
 
-	errs := []error{}
 	for i := range podList.Items {
 		pod := &podList.Items[i]
 		mutateFn := func(pod *corev1.Pod) error {
--- a/internal/controller/controller/controller_sync_status_test.go
+++ b/internal/controller/controller/controller_sync_status_test.go
@@ -131,13 +131,22 @@ func TestControllerReconciler_syncHAStatus(t *testing.T) {
 			},
 		},
 		{
-			name:       "Slurm error is returned",
+			// Regression: a transport error used to abort before labeling, which
+			// left the Controller Service without endpoints. A Service without
+			// endpoints is what makes Slurm unreachable, so the next reconcile
+			// failed the same way and nothing ever relabeled a pod.
+			name:       "Slurm error defaults to primary",
 			controller: newController(false),
 			pods: []*corev1.Pod{
-				newPod(newController(false), 0, true),
+				newPod(newController(false), 0, false),
+				newPod(newController(false), 1, true),
 			},
 			pingErr: errors.New("internal error"),
 			wantErr: true,
+			wantActive: map[string]bool{
+				"slurm-controller-0": true,
+				"slurm-controller-1": false,
+			},
 		},
 	}
 	for _, tt := range tests {
@@ -158,9 +167,12 @@ func TestControllerReconciler_syncHAStatus(t *testing.T) {
 			err := r.syncHAStatus(t.Context(), tt.controller)
 			if tt.wantErr {
 				require.Error(t, err)
-				return
+			} else {
+				require.NoError(t, err)
 			}
-			require.NoError(t, err)
+
+			// Labels are reconciled either way: surfacing the Slurm failure
+			// must not cost the Service its endpoints.
 
 			got := make([]*corev1.Pod, 0, len(tt.pods))
 			for _, pod := range tt.pods {


internal/controller/controller/controller_sync_status.go
```diff
@@ -104,8 +104,22 @@ func (r *ControllerReconciler) syncHAStatus(
...
+		// active. The Controller Service selects pods by LabelControllerActive,
+		// so returning here leaves the Service without endpoints — and a
+		// Service without endpoints is exactly what makes Slurm unreachable,
+		// so every later reconcile fails the same way and no pod is ever
+		// labeled again. The deadlock survives operator restarts because the
+		// Slurm client is created once and kept, so the call keeps failing
+		// with a transport error rather than ErrNoSlurmClient. Treat any
+		// failure like a missing client: fall through with no pings, label the
+		// primary, and let a later reconcile move the label once Slurm answers.
+		if !errors.Is(err, slurmcontrol.ErrNoSlurmClient) {
```
vivian-hafener: Errors raised here are effectively suppressed - even though they are logged, they do not make it into the slice of errors that is used to compose the aggregate at the end of this function. I would prefer it if the slice of error `errs` were to be declared earlier, and this error appended to it.

internal/controller/controller/controller_sync_status.go
```diff
@@ -104,8 +104,22 @@ func (r *ControllerReconciler) syncHAStatus(
 	}
 
 	pings, err := r.slurmControl.GetActiveHAController(ctx, controller)
-	if err != nil && !errors.Is(err, slurmcontrol.ErrNoSlurmClient) {
-		return err
+	if err != nil {
+		// Do not bail out when Slurm cannot be asked which controller is
```
vivian-hafener: Comment is overly verbose. The first sentence alone is fine.

internal/controller/controller/controller_sync_status.go
```diff
@@ -104,8 +104,22 @@ func (r *ControllerReconciler) syncHAStatus(
...
+		// labeled again. The deadlock survives operator restarts because the
+		// Slurm client is created once and kept, so the call keeps failing
+		// with a transport error rather than ErrNoSlurmClient. Treat any
+		// failure like a missing client: fall through with no pings, label the
+		// primary, and let a later reconcile move the label once Slurm answers.
+		if !errors.Is(err, slurmcontrol.ErrNoSlurmClient) {
+			log.FromContext(ctx).V(1).Info("Failed to determine the active controller, defaulting to the primary",
+				"Controller", klog.KObj(controller), "error", err)
+		}
+		pings = nil
```
vivian-hafener: I believe that is is unecessary to set pings to nil here - all codepaths in `GetActiveHAController()` that return a non-nil error return a nil value for pings.

Removing this means that you can also retain the original structure of:
```
if err != nil && !errors.Is(err, slurmcontrol.ErrNoSlurmClient) {
   ...
}
```

vivian-hafener (changes requested):
I've added comments where improvements could be made. In addition, your commit should include `Signed-off-by:` and `Changelog:` trailers following the repository convention.

AndreyZa:
Thanks for the review — all four points addressed in the amended commit.

* **Suppressed error.** `errs` is now declared before the `GetActiveHAController` call and the failure is appended to it, so it ends up in the aggregate instead of only being logged. The labels are still reconciled first, which is the point of the fix: the caller learns about the Slurm failure, but not at the price of the Service losing its endpoints.
* **Nil assignment.** Dropped, and with it the restructuring — the condition is back to the original `if err != nil && !errors.Is(err, slurmcontrol.ErrNoSlurmClient)`.
* **Comment.** Trimmed to the first sentence.
* **Trailers.** `Signed-off-by:` and `Changelog:` added; sorry for missing them.

One consequence worth flagging: `syncHAStatus` now returns an error in the "Slurm unreachable" case where the previous revision returned nil, so the regression test needed updating. It now asserts both halves of the contract — the error is surfaced **and** the labels are reconciled — rather than returning early on error. `go build ./...`, `gofmt` and `go test ./internal/controller/controller/...` are clean.

vivian-hafener:
I'll take a look, thanks.

vivian-hafener:
See: https://github.com/SlinkyProject/slurm-operator/issues/257#issuecomment-5590148391

AndreyZa:
Closing: the premise of this PR was wrong — see https://github.com/SlinkyProject/slurm-operator/issues/257#issuecomment-5590148391 and my reply there. The code I analysed lives only on `main`, and the deadlock we observed was caused by a stale Service selector left behind by a `main`-built operator on a test cluster, not by 1.2.2. I cannot demonstrate the failure this change guards against, so it should not take up review time. Happy to reopen if you consider it useful as a hardening on its own.

PR #260, opened 2026-09-06
fix(nodeset): give DaemonSet-mode pods the standard daemon tolerations

### LINKED ISSUE #259: [Bug]: NodeSet pod is never recreated on a cordoned node in DaemonSet scaling mode
### What happened

With `scalingMode: DaemonSet`, a NodeSet pod cannot be **created** on a cordoned node. Cordoning adds `node.kubernetes.io/unschedulable:NoSchedule`; the pod built by `newNodeSetPodDaemon` does not tolerate it, so `PodShouldRunOnNode` returns `shouldRun=false` and the node drops out of `status.desired`.

Existing pods are unaffected, and that is exactly what hides the problem: cordon does not evict, so a cordoned node keeps its pod and everything looks healthy. The gap only appears once a pod has to come **back** — a rolling update after an image change is the ordinary way to get there. From then on the node stays without a pod for as long as it is cordoned, and nothing in the reconcile loop recovers it.

This also disagrees with the rest of the controller. `syncCordon` treats "pod on a cordoned node" as a legitimate state and drains the matching Slurm node for it, and `PodShouldRunOnNode` already returns `shouldContinueRunning=true` for that case. Only the create path takes the opposite view.

### Why it matters in practice

We run compute nodes under a power policy that cordons a node before powering it off. A slurmd image bump rolled the NodeSet, the sleeping nodes lost their pods, and they could never get them back: the acceptance gate waits for the slurmd pod, never sees it, and returns the node to quarantine — still cordoned. Four nodes were stuck that way until we added the toleration by hand.

### What did you expect to happen

The same behaviour as the upstream DaemonSet controller, which applies a fixed set of tolerations to its own pods (`AddOrUpdateDaemonPodTolerations`), `node.kubernetes.io/unschedulable:NoSchedule` among them.

### How can we reproduce it?

Operator and chart 1.2.2, single-node k0s cluster.

```console
$ kubectl -n slurm patch nodesets.slinky.slurm.net slurm-worker-lab \
    --type=merge -p '{"spec":{"scalingMode":"DaemonSet"}}'
$ kubectl -n slurm get nodesets.slinky.slurm.net slurm-worker-lab \
    -o custom-columns=DESIRED:.status.desired,REPLICAS:.status.replicas --no-headers
1     1

$ kubectl cordon <node>          # pod survives, as expected
$ kubectl -n slurm delete pod <slurmd pod>

$ kubectl -n slurm get pods -l app.kubernetes.io/name=slurmd
No resources found in slurm namespace.
$ kubectl -n slurm get nodesets.slinky.slurm.net slurm-worker-lab \
    -o custom-columns=DESIRED:.status.desired,REPLICAS:.status.replicas --no-headers
<none>   <none>
```

The node is `Ready` and healthy; the NodeSet simply has nothing to place. Adding the toleration by hand brings it straight back:

```console
$ kubectl -n slurm patch nodesets.slinky.slurm.net slurm-worker-lab --type=json \
    -p '[{"op":"add","path":"/spec/template/spec/tolerations/-","value":
         {"key":"node.kubernetes.io/unschedulable","operator":"Exists","effect":"NoSchedule"}}]'
$ kubectl -n slurm get pods -l app.kubernetes.io/name=slurmd --no-headers
slurm-worker-lab-mdkjr   2/2   Running   0   20s
$ kubectl -n slurm get nodesets.slinky.slurm.net slurm-worker-lab \
    -o custom-columns=DESIRED:.status.desired,REPLICAS:.status.replicas --no-headers
1     1
```

### Version

1.2.2

### Anything else we need to know?

Patch in the linked PR: apply `AddOrUpdateDaemonPodTolerations` to both the real DaemonSet-mode pod and the simulated pod used for predicate evaluation. The simulated one matters too — the predicate has to see the same tolerations the real pod gets, otherwise it still reports "should not run".

AndreyZa:
## Summary

In `scalingMode: DaemonSet` a NodeSet pod cannot be created on a cordoned node. Cordon adds `node.kubernetes.io/unschedulable:NoSchedule`, the pod built by `newNodeSetPodDaemon` does not tolerate it, `PodShouldRunOnNode` returns `shouldRun=false`, and the node leaves `status.desired`.

Existing pods are unaffected — cordon does not evict — so nothing looks wrong until a pod has to come back. A rolling update after an image change is the ordinary way to get there, and from then on the node stays without a pod for as long as it is cordoned.

It is also inconsistent with the neighbouring code: `syncCordon` treats "pod on a cordoned node" as a legitimate state and drains the matching Slurm node for it, and `PodShouldRunOnNode` already returns `shouldContinueRunning=true` for it. Only the create path disagrees.

The fix applies `daemonutils.AddOrUpdateDaemonPodTolerations` — the same helper the upstream DaemonSet controller uses for its own pods — in two places:

* `newNodeSetPodDaemon`, the real DaemonSet-mode pod;
* `newSimulatedDaemonPod`, the pod used for predicate evaluation. This one matters too: the predicate has to see the same tolerations the real pod gets, otherwise it still reports "should not run" and the fix has no effect.

The package is already imported (`daemonutils "k8s.io/kubernetes/pkg/controller/daemon/util"`), so the change is two calls. `StatefulSet` scaling mode is untouched.

Closes #259.

## Checklist

- [x] I have read the
  and the
- [x] New or existing tests cover these changes (where applicable). (Existing `go test ./internal/controller/nodeset/...` passes; the behaviour was verified on a cluster, see below.)
- [x] Documentation is updated if user-visible behavior changes. (No documentation change: DaemonSet-mode pods now get the tolerations the Kubernetes DaemonSet controller documents for daemon pods.)

## Breaking Changes

None. Pods gain the standard daemon tolerations (`unschedulable`, `not-ready`, `unreachable`, `disk-pressure`, `memory-pressure`, `pid-pressure`, `network-unavailable`), exactly as pods of a Kubernetes `DaemonSet` do. A cordoned node keeps its slurmd pod, which is what `syncCordon` already assumes.

## Testing Notes

Reproduced and verified on operator + chart 1.2.2, single-node cluster:

| step | before | after |
|---|---|---|
| DaemonSet mode, node uncordoned | desired=1, pod Running | same |
| cordon the node | pod survives | same |
| delete the pod | **desired=0, no pod, never recovers** | desired=1, pod Running on the cordoned node |

`go build ./...` clean, `gofmt` clean, `go test ./internal/controller/nodeset/...` passes.

## Additional Context

Found on a cluster that cordons nodes before powering them down: a slurmd image bump rolled the NodeSet, and the cordoned nodes never got their pods back — the NodeSet simply had nothing to place on them for as long as they stayed cordoned.

### DIFF
--- a/internal/controller/nodeset/nodeset_sync.go
+++ b/internal/controller/nodeset/nodeset_sync.go
@@ -1300,6 +1300,10 @@ func (r *NodeSetReconciler) newNodeSetPodDaemon(
 	hostnameOverride := node.Annotations[slinkyv1beta1.AnnotationNodeHostnameOverride]
 
 	pod := nodesetutils.NewNodeSetDaemonSetPod(client, nodeset, controller, nodeName, hostnameOverride, revisionHash)
+	// Same tolerations the DaemonSet controller gives its own pods. Without
+	// them a NodeSet pod cannot be (re)created on a cordoned node, because
+	// cordon adds node.kubernetes.io/unschedulable:NoSchedule.
+	daemonutils.AddOrUpdateDaemonPodTolerations(&pod.Spec)
 	return pod, nil
 }
 
@@ -1325,6 +1329,10 @@ func newSimulatedDaemonPod(
 	}
 
 	pod := nodesetutils.NewNodeSetSimulatedPod(client, nodeset, controller, nodeName)
+	// The predicate must see the same tolerations the real pod gets,
+	// otherwise PodShouldRunOnNode reports "should not run" for nodes the
+	// pod would in fact tolerate.
+	daemonutils.AddOrUpdateDaemonPodTolerations(&pod.Spec)
 	return pod, nil
 }
 


AndreyZa:
Force-pushed `3e93da3` → `dc33990`: commit message only, the tree is byte-identical.

The `Changelog` trailer was a single 159-character line, over `body-max-line-length`, so the repo's commitlint hook rejected it; it is now folded across continuation lines, which `git interpret-trailers --parse` still reads as one trailer. Also replaced two em-dashes with ASCII, since upstream commit messages carry none.

No change to the patch itself: `go build ./...` clean, `gofmt` clean, `go test ./internal/controller/nodeset/...` passes.

Issue #119, opened 2026-01-26
[Bug]: sssd config passed at helm values not translated to loginset manifest


iam-mg-Harish:
## Description

I tried passing my custom sssd configuration to both sssd.conf and sssd.secretRef, but it is not translated into sssdConfRef spec of LoginSet CR after deployment.

## Steps to Reproduce

1. Pass custom sssd configuration in sssd section of slurm's helm values file.
2. Deploy the slurm release with the values file.
3. Check whether the custom sssd config is reflected in the loginset

klairecodes:
Please provide your full `sssd.config`, `values.yaml`, and the outputs of `kubectl -n slurm get loginsets.slinky.slurm.net -o yaml` and `kubectl -n slurm describe secrets slurm-login-slinky-sssd-conf`.

iam-mg-Harish:
**Loginset**
```
apiVersion: v1
items:
- apiVersion: slinky.slurm.net/v1beta1
  kind: LoginSet
  metadata:
    annotations:
      meta.helm.sh/release-name: slurm
      meta.helm.sh/release-namespace: slurm
    creationTimestamp: "2026-01-27T17:24:52Z"
    generation: 1
    labels:
      app.kubernetes.io/managed-by: Helm
      app.kubernetes.io/part-of: slurm
      app.kubernetes.io/version: "25.11"
      helm.sh/chart: slurm-1.0.1
    name: slurm-login-slinky
    namespace: slurm
    resourceVersion: "28368812"
    uid: 081d3b7c-c3d6-47b1-9a11-d16b501f9666
  spec:
    controllerRef:
      name: slurm
      namespace: slurm
    initconf:
      image: alpine:latest
      imagePullPolicy: Always
      resources: {}
    login:
      env: []
      image: repository: ghcr.io/slinkyproject/login:25.11-ubuntu24.04
      imagePullPolicy: Always
      resources: {}
      securityContext:
        privileged: false
      volumeMounts:
      - mountPath: /apps
        name: pvc
        subPath: apps
      - mountPath: /home
        name: pvc
        subPath: home
    replicas: 1
    service:
      metadata: {}
      spec:
        type: LoadBalancer
    sssdConfRef:
      key: sssd.conf
      name: slurm-login-slinky-sssd-conf
    template:
      spec:
        affinity: {}
        imagePullSecrets: null
        initContainers: []
        nodeSelector:
          kubernetes.io/os: linux
        priorityClassName: null
        resources: {}
[...]

PR #263, opened 2026-09-11
feat(helm): allow a per-set sssd.conf for LoginSet and NodeSet

AndreyZa:
## Summary

`sssdConfRef` is a per-set field on both `LoginSet` and `NodeSet`, but the chart can only fill it from the cluster-wide value: `slurm.sssdConf.name` reads `.Values.sssd.secretRef` and never sees the set it is rendering. Every LoginSet and NodeSet in a release therefore resolves identities from the same directory.

That becomes limiting as soon as one release has to serve two directories: two LoginSets, or a LoginSet and a NodeSet, each resolving identities from its own domain. A single shared `sssd.conf` gives every set every domain, which is exactly what such a split exists to prevent.

The only workaround today is to patch `sssdConfRef` on the CR after install. That patch survives an ordinary `helm upgrade` (the field is identical in the old and new rendered manifests, so it is not part of the patch Helm computes), but it is silently lost when the global `sssd.secretRef` changes, when the CR is recreated, or when the release is reinstalled — and the failure mode is a login where users simply "do not exist".

`loginsets.<name>.sssd.secretRef` and `nodesets.<name>.sssd.secretRef` now replace the cluster-wide reference **as a whole**: a set that names its own secret takes `key` from that same block, defaulting to `sssd.conf`, never inheriting the cluster-wide `key`. Composing the reference out of one field from each source would be a quiet way to read the wrong key of the right secret, so a `key` without a `name` fails the render rather than half-resolving. A set that says nothing keeps the cluster-wide value. Four small edits:

* `templates/loginset/loginset-cr.yaml` — use the per-set reference when it names a secret, otherwise the existing helpers;
* `templates/nodeset/nodeset-cr.yaml` — same, before `ssh.sssdConfRef` is set;
* `values.yaml` — the new key documented under `loginsetDefaults` and `nodesetDefaults`, plus the regenerated README;
* `tests/loginset_test.yaml`, `tests/nodeset_test.yaml` — six `helm unittest` cases per template.

The helpers themselves are untouched, so the cluster-wide value keeps working exactly as before.

9c6d391 made `sssd.conf` cluster-wide on purpose, in preparation for NodeSet support, and this does not walk that back: the cluster-wide value stays the default and remains the only value a release needs. The per-set key is an optional override, and with no per-set value the rendered `sssdConfRef` is byte-for-byte what it was. Related: #119 wired `sssd.secretRef` through to the CR in the first place; this makes that wiring per-set.

## Checklist

- [x] I have read the [CONTRIBUTING.md](https://github.com/SlinkyProject/slurm-operator/blob/main/CONTRIBUTING.md) and the [Code of Conduct](https://github.com/SlinkyProject/slurm-operator/blob/main/CODE_OF_CONDUCT.md).
- [x] New or existing tests cover these changes (where applicable). (Twelve `helm unittest` cases, six per template, in `helm/slurm/tests/loginset_test.yaml` and `helm/slurm/tests/nodeset_test.yaml`.)
- [x] Documentation is updated if user-visible behavior changes. (`values.yaml` comments and the helm-docs README.)

## Breaking Changes

None. With no per-set value present the rendered `sssdConfRef` is byte-for-byte what it was, and a set that carries no `sssd` key at all — or an explicit `sssd: null` — renders the cluster-wide value, as it did before.

## Testing Notes

`make helm-unittest` — 16 suites, 188 tests, all passing (176 on `main`, so twelve new cases). Six per template:

| case | `sssdConfRef` |
|---|---|
| set declares no `sssd` | cluster-wide name and key |
| set names its own secret | that name, key `sssd.conf` |
| set names a secret and a key | both from the set |
| set gives a `key` with no `name` | render fails with a named error |
| set clears `sssd` with `null` | cluster-wide name and key |
| two sets, one overriding | each its own |

Both directions were checked, not just the green run:

* with `templates/loginset/loginset-cr.yaml` and `templates/nodeset/nodeset-cr.yaml` reverted to `main`, the overriding cases fail and the rest pass;
* an earlier revision of this patch reached the per-set value with `dig`, which panics on `sssd: null` (`error calling dig: interface conversion: interface {} is nil`) where `main` renders fine. The two `null` cases fail against that revision and pass against this one, which reads the value the way the rest of the chart does (`| default dict`).

Defaults unchanged: rendering the LoginSet and NodeSet manifests of a release that sets no per-set value gives output identical to `main`. `helm lint helm/slurm` passes, `make helm-docs` produces no further diff.

## Additional Context

Found when a second LoginSet added to an existing release had to resolve identities from a different directory than the first: it came up with the first one's domain, and users of the second directory were unresolvable with no error anywhere — the pod is healthy, `getent` simply returns nothing.

Two things adjacent to this diff, neither addressed here, both happy to take to separate PRs:

* `docs/installation.md` still documents `--set-file "loginsets.slinky.sssdConf=${HOME}/sssd.conf"`, a key 9c6d391 removed. `--set-file` on an unknown key is silently a no-op, so that instruction now quietly leaves the default `sssd.conf` in place.
* Rotating a per-set `sssd.conf` will not reach running NodeSet pods promptly. `internal/controller/loginset/eventhandler/eventhandler_secret.go` compares an incoming Secret against `loginset.SssdSecretKey()`, but the NodeSet handler matches only the Slurm and JWT keys, so the change waits for an unrelated reconcile. Same on `main`; not a regression from this PR.


vivian-hafener (changes requested):
Please add Helm unit tests to cover the added behavior. Also, the commit does not include the project's required git trailers.

AndreyZa:
Both points addressed — force-pushed (`49901ae` → `a548d25`).

**Trailers.** The commit now carries `Signed-off-by` and `Changelog`, in one contiguous final block. The `Changelog` line is folded across continuation lines: written as a single line it ran to 206 characters and the commitlint hook rejected it on `body-max-line-length`. `git interpret-trailers --parse` still reads it as one trailer, and `commitlint --strict` exits 0.

**Helm unit tests.** Five cases per template, in `tests/loginset_test.yaml` and `tests/nodeset_test.yaml` — the suite goes 176 → 186:

| case | expected `sssdConfRef.name` |
|---|---|
| set declares no `sssd` | cluster-wide value |
| set overrides `name` and `key` | per-set value |
| set overrides `name` only | per-set name, cluster-wide key |
| set clears `sssd` with `null` | cluster-wide value |
| two sets, one overriding | each its own |

I checked they actually fail without the change, not just that they pass with it: with both templates reverted to `main`, the six overriding cases fail and the rest pass.

**Writing them turned up a defect in the first revision.** It reached the per-set value with `dig "sssd" "secretRef" "name" "" $loginset`, and `dig` panics when the key is present but null:

```
$ helm template rel helm/slurm --set 'loginsets.slinky.sssd=null' ...
Error: template: slurm/templates/loginset/loginset-cr.yaml:36:13:
  executing ... at <dig "sssd" "secretRef" "name" "" $loginset>:
  error calling dig: interface conversion: interface {} is nil, not map[string]interface {}
```

`main` renders that same values file fine, so the first revision would have broken a release that clears the key. It now reads the value the way the rest of the chart does — `($loginset.sssd | default dict).secretRef | default dict` — and the two `null` cases in the table above are there to keep it that way. `helm lint` passes and `make helm-docs` produces no further diff; the helm-docs type annotation also moved onto the `secretRef` leaf, matching how the cluster-wide `sssd.secretRef` is annotated.

AndreyZa:
Revised the fallback contract before you get to this — force-pushed `7976afc` → `a53adc4`. Flagging it because my earlier comment described the old behaviour.

It used to fall back **field by field**: a set that named its own secret but no key inherited `key` from the cluster-wide `sssd.secretRef`. That composes a reference out of two sources, and the result is the wrong kind of quiet: with a cluster-wide `key: bind-password`, a set asking only for `name: team-a-secret` rendered `name: team-a-secret, key: bind-password` — the right secret read at the wrong key, which either wedges the pod on mount or hands sssd a file that is not an `sssd.conf`.

Now the per-set block replaces the cluster-wide reference as a whole. A set that names a secret takes `key` from the same block, defaulting to `sssd.conf`; the cluster-wide `key` is never inherited. A `key` with no `name` is a half-configured reference, so it fails the render with the set named, rather than silently resolving to the cluster-wide directory:

```
Error: execution error at (slurm/templates/loginset/loginset-cr.yaml:37:8):
  loginset `students` sets sssd.secretRef.key without sssd.secretRef.name.
```

`fail` in a template felt heavy-handed until I looked at what the alternative costs: the whole point of the change is that a set resolves identities from the directory the operator meant, and every silent path here ends with a healthy pod serving the wrong population. It matches how the chart already guards partitions in `controller/_helpers.tpl`.

Tests go 176 → 188, six cases per template; the two new ones cover the key-without-name failure and the key defaulting rather than inheriting. The description and the table are updated.

